PR #26, opened 2026-04-02
fix: resolve ${VAR} brace syntax in hook command paths

### LINKED ISSUE #17: Bug: hook registration writes unresolved env vars to settings.json
## Problem

When `arc install` or `arc upgrade` registers hooks from a package's `pai-manifest.yaml`, it writes the hook command verbatim — including `${MINER_DIR}` or similar env var references. These vars are not resolved to absolute paths, so the hooks break if the var isn't set in the shell environment.

This has caused repeated failures: the Miner package hooks reference `${MINER_DIR}/src/hooks/MinerEventLogger.hook.ts`, which resolves to `/src/hooks/MinerEventLogger.hook.ts` when `MINER_DIR` is unset. The fix (manually replacing with the absolute path) gets overwritten on every `arc upgrade`.

**This has been manually fixed at least twice now and keeps regressing.**

## Expected Behavior

`arc install`/`arc upgrade` should resolve hook paths to absolute paths when writing to `~/.claude/settings.json`. Since arc already knows the install location (`~/.config/arc/pkg/repos/<name>/`), it should substitute that path at install time rather than preserving template variables.

## Current Behavior

```json
// What arc writes (broken):
"command": "bun ${MINER_DIR}/src/hooks/MinerEventLogger.hook.ts"

// What it should write:
"command": "bun /Users/fischer/.config/arc/pkg/repos/miner/src/hooks/MinerEventLogger.hook.ts"
```

## Where to Fix

The `registerHooks` function in arc's install/upgrade flow — it reads hook commands from `pai-manifest.yaml` and writes them to settings.json without resolving paths.

## Impact

Every hook event (UserPromptSubmit, PreToolUse, PostToolUse, Stop) throws an error on every Claude Code interaction. 4 errors per tool call across all sessions.

jcfischer:
## Summary
- Hook path resolution only matched `$PKG_DIR` / `$NAME_DIR` but not `${PKG_DIR}` / `${NAME_DIR}`
- Since manifests use the brace syntax, hooks were written with unresolved env vars, breaking every session
- Updated regex patterns in `resolveCommandPaths` to accept both forms

## Test plan
- [x] Added tests for `${PKG_DIR}` brace syntax resolution
- [x] Added tests for `${NAME_DIR}` brace syntax resolution
- [x] All 231 existing tests pass

Closes #17

### DIFF
--- a/src/lib/hooks.ts
+++ b/src/lib/hooks.ts
@@ -108,7 +108,8 @@ export function resolveHooksFromManifest(
 }
 
 /**
- * Replace $PKG_DIR and $<NAME>_DIR in hook commands with the actual install path.
+ * Replace $PKG_DIR / ${PKG_DIR} and $<NAME>_DIR / ${<NAME>_DIR}
+ * in hook commands with the actual install path.
  */
 function resolveCommandPaths(
   hooks: InlineHook[],
@@ -116,12 +117,12 @@ function resolveCommandPaths(
   packageName: string,
 ): InlineHook[] {
   const nameUpper = packageName.toUpperCase().replace(/-/g, "_");
-  const namePattern = new RegExp(`\\$${nameUpper}_DIR`, "g");
+  const namePattern = new RegExp(`\\$\\{?${nameUpper}_DIR\\}?`, "g");
 
   return hooks.map((hook) => ({
     ...hook,
     command: hook.command
-      .replace(/\$PKG_DIR/g, installPath)
+      .replace(/\$\{?PKG_DIR\}?/g, installPath)
       .replace(namePattern, installPath),
   }));
 }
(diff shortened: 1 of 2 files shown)


mellanon (approved):
Clean, correct fix. The regex changes from `$PKG_DIR` to `$\{?PKG_DIR\}?` (and same for NAME_DIR) correctly handle both syntaxes — with and without braces — which matches how shell variable syntax works in manifests.

**Verified:**
- The fix is surgical: 2 regex patterns updated, JSDoc comment updated, nothing else touched
- Both new tests cover the exact failure case from issue #17 (brace syntax not resolving)
- The regex `\{?` / `\}?` makes the braces optional, so the old bare `$PKG_DIR` form still works — no regression
- Tested against the real-world case: `${MINER_DIR}/src/hooks/MinerEventLogger.hook.ts` resolving to the absolute install path

Ship it.

PR #38, opened 2026-04-03
feat: arc upgrade --force

### LINKED ISSUE #13: arc upgrade --force: bypass version check and force upgrade
## Feature Request

`arc upgrade` should support a `--force` flag that bypasses the version number check and forces the upgrade to run regardless.

## Use Case

When debugging upgrade issues or re-applying an upgrade after a failed attempt, the current version check blocks the operation. A `--force` option would allow re-running the upgrade pipeline without manually manipulating version state.

## Proposed Behavior

```bash
arc upgrade              # Current: skips if already on latest version
arc upgrade --force      # New: runs upgrade pipeline regardless of version
```

mellanon:
## Summary

- Add `--force` flag to `arc upgrade` that bypasses the version check and re-runs the full upgrade pipeline (scripts, hooks, symlinks, capabilities) even when already at the latest version
- Passes `force` through all upgrade paths: single package, library upgrade, and upgrade-all
- Displays `force-upgraded at X.Y.Z` instead of `already at X.Y.Z` when force is used

## Test plan

- [x] New test: "with force flag, re-runs upgrade even when at latest version" -- verifies `upgradePackage()` runs the full pipeline with `{ force: true }` and returns success with matching oldVersion/newVersion
- [x] All 268 existing tests pass (no regressions)
- [x] Type-check passes (`tsc --noEmit`)

closes #13

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -311,7 +311,7 @@ program
         console.log(formatUpgradeResults(results, { force: true }));
       }
     } else {
-      // Check for library:artifact syntax
+      // name is guaranteed non-null here — commander validates required args for single-package paths
       const libRef = parseLibraryRef(name!);
       let upgradeName = libRef?.artifactName ?? name!;
       let isLibraryUpgrade = false;
--- a/src/commands/upgrade.ts
+++ b/src/commands/upgrade.ts
@@ -303,21 +303,30 @@ export async function upgradePackage(
 
 /**
  * Upgrade all installed packages that have newer versions.
+ * When force=true, skips the expensive checkUpgrades (git fetch + registry
+ * lookup per package) and instead gets all active packages directly from the DB.
  */
 export async function upgradeAll(
   db: Database,
   paths: PaiPaths,
   opts?: { force?: boolean }
 ): Promise<UpgradeResult[]> {
-  const checks = await checkUpgrades(db, paths);
-  const upgradable = opts?.force
-    ? checks // force: attempt all installed packages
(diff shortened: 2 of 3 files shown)


src/cli.ts
```diff
@@ -301,29 +302,41 @@ program
...
+        console.log("No packages installed.");
+      } else {
+        console.log(formatUpgradeResults(results, { force: true }));
+      }
     } else {
       // Check for library:artifact syntax
-      const libRef = parseLibraryRef(name);
-      let upgradeName = libRef?.artifactName ?? name;
+      const libRef = parseLibraryRef(name!);
+      let upgradeName = libRef?.artifactName ?? name!;
```
mellanon: **[nit/CodeQuality]** Non-null assertions (`name!`) are safe here but undocumented.

The 4 `name!` assertions are logically safe because the preceding `if` and `else if` branches handle all `!name` cases, guaranteeing `name` is defined in this `else` branch. Per project convention, consider adding a brief comment or restructuring to avoid the assertions — e.g., narrowing `name` with a type guard at the top of the branch.

mellanon (commented):
Full 5-lens + duplication review complete. No critical findings. 1 warning (missing test coverage for force flag on batch/library upgrade paths), 3 suggestions, 1 nit.

The core implementation is sound — force flag correctly bypasses the version gate while preserving the full upgrade pipeline.

**Verdict: approve** (submitted as comment because GitHub blocks self-approval).

PR #55, opened 2026-04-10
feat: add metafactory source type (A-401 F-1)

jcfischer:
## Summary
- Adds `type` field to `RegistrySource`: `"registry"` (default, REGISTRY.yaml) or `"metafactory"` (API-based)
- Adds `token` field for Bearer authentication (populated by device code auth in F-2)
- `arc source add` gains `--type` flag with validation
- Default source changes from GitHub REGISTRY.yaml to `https://meta-factory.ai` with `type: metafactory`, `tier: official`
- `formatSourceList` shows type, never exposes tokens

## Validation rules for metafactory sources
- URL must be HTTPS
- URL must not end in `.yaml`/`.yml` (suggests `--type registry` instead)
- Invalid type values get clear error with valid options listed

## Backward compatibility
- Sources without `type` default to `"registry"` -- existing sources.yaml files work unchanged
- Unknown fields silently ignored (forward compat)

## Tests
17 new tests added. Full suite: **318/318 passing**.

| Test Group | Count | Covers |
|-----------|-------|--------|
| validateSource | 8 | Type parsing, URL rules, error messages |
| getSourceType | 3 | Explicit type, default fallback |
| Token handling | 2 | Parse from YAML, empty = falsy |
| Backward compat | 1 | Legacy sources.yaml without type |
| Forward compat | 1 | Unknown fields ignored |
| Display security | 1 | Token never in formatSourceList output |
| addSource integration | 1 | Validation wired into addSource |

## SpecFlow
This is F-1 of a 5-feature decomposition of A-401 (metafactory registry as first-class source):
- **F-1: Source type** (this PR)
- F-2: Device code auth flow
- F-3: Registry API client
- F-4: Install from registry
- F-5: Search across sources

## Test plan
- [ ] `bun test` -- 318/318 green
- [ ] `bunx tsc --noEmit` clean
- [ ] `arc source add test https://meta-factory.ai --type metafactory` succeeds
- [ ] `arc source add test http://x.com --type metafactory` fails with HTTPS error
- [ ] `arc source add test https://x.com/R.yaml --type metafactory` suggests --type registry
- [ ] `arc source list` shows type column, no tokens
- [ ] Delete sources.yaml, run `arc source list` -- creates default with type: metafactory, tier: official

@the-metafactory/luna — requesting review

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -34,7 +34,7 @@ import {
   formatCatalogList,
   formatCatalogSearch,
 } from "./commands/catalog.js";
-import type { CatalogEntry, ArtifactType, PackageTier } from "./types.js";
+import type { CatalogEntry, ArtifactType, PackageTier, RegistrySource, SourceType } from "./types.js";
 import { loadCatalog, saveCatalog, findEntry } from "./lib/catalog.js";
 import {
   loadSources,
@@ -42,6 +42,7 @@ import {
   addSource,
   removeSource,
   formatSourceList,
+  validateSource,
 } from "./lib/sources.js";
 import {
   searchAllSources,
@@ -498,25 +499,29 @@ source
   .command("add <name> <url>")
   .description("Add a new registry source")
   .option("-t, --tier <tier>", "Trust tier (official|community|custom)", "community")
-  .action(async (name: string, url: string, opts: { tier: string }) => {
+  .option("--type <type>", "Source type (registry|metafactory)", "registry")
+  .action(async (name: string, url: string, opts: { tier: string; type: string }) => {
     const paths = createPaths();
     const config = await loadSources(paths.sourcesPath);
 
-    // Validate URL format
-    if (!url.startsWith("https://") && !url.startsWith("http://") && !url.startsWith("file://")) {
-      console.error(`Error: Invalid URL "${url}". Must start with https://, http://, or file://`);
+    const newSource: RegistrySource = {
+      name,
+      url,
+      tier: opts.tier as PackageTier,
+      enabled: true,
+      ...(opts.type !== "registry" ? { type: opts.type as SourceType } : {}),
+    };
+
+    const validation = validateSource(newSource);
+    if (!validation.valid) {
+      console.error(`Error: ${validation.error}`);
       process.exit(1);
     }
 
     try {
-      addSource(config, {
-        name,
-        url,
-        tier: opts.tier as PackageTier,
-        enabled: true,
-      });
+      addSource(config, newSource);
       await saveSources(paths.sourcesPath, config);
-      console.log(`Added source "${name}" [${opts.tier}]`);
+      console.log(`Added source "${name}" [${opts.tier}] (${opts.type})`);
     } catch (err: any) {
       console.error(`Error: ${err.message}`);
       process.exit(1);
--- a/src/lib/remote-registry.ts
+++ b/src/lib/remote-registry.ts
@@ -44,6 +44,12 @@ export async function fetchRemoteRegistry(
   cachePath: string,
   forceRefresh?: boolean
 ): Promise<RegistryConfig | null> {
+  // metafactory API sources are handled by a dedicated client (F-3).
+  // Skip gracefully here so arc search/update don't choke on HTML responses.
+  if (source.type === "metafactory") {
+    return null;
+  }
+
   // Local file source — read directly, no caching
   if (source.url.startsWith("file://")) {
     const filePath = source.url.replace("file://", "");
--- a/src/lib/sources.ts
+++ b/src/lib/sources.ts
@@ -1,12 +1,16 @@
 import { readFile, writeFile } from "fs/promises";
 import { existsSync } from "fs";
 import YAML from "yaml";
-import type { SourcesConfig, RegistrySource, PackageTier } from "../types.js";
+import type { SourcesConfig, RegistrySource, SourceType } from "../types.js";
+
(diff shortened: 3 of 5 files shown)


mellanon (commented):
## Review: A-401 F-1 — metafactory source type

Clean foundation PR. Good type design, excellent test coverage, proper token security. Two findings — one blocking, one important for the transition window.

---

### Finding 1 — Registry-type URL validation dropped (blocking)

The old `source add` code validated _all_ URLs against `https://`, `http://`, `file://`:

```js
// OLD (cli.ts:506)
if (!url.startsWith("https://") && !url.startsWith("http://") && !url.startsWith("file://")) {
```

The new code delegates to `validateSource()`, which only validates metafactory URLs. For `registry` type, there is now **zero URL validation**:

```bash
# This now succeeds (shouldn't):
arc source add foo garbage://nonsense --type registry
```

**Fix:** Add basic URL scheme validation for registry type in `validateSource()`:
```ts
if (type === "registry") {
  if (!source.url.startsWith("https://") && !source.url.startsWith("http://") && !source.url.startsWith("file://")) {
    return { valid: false, error: `Invalid URL "${source.url}". Must start with https://, http://, or file://` };
  }
}
```

---

### Finding 2 — Default source broken during F-1→F-3 window (non-blocking, worth discussing)

The default source changed from the GitHub REGISTRY.yaml URL to `https://meta-factory.ai` with type `metafactory`. But `fetchRemoteRegistry()` (remote-registry.ts:42-142) doesn't understand source types — it will:

1. Fetch `https://meta-factory.ai` (gets HTML)
2. `YAML.parse(html)` → no `.registry` key → return `null`
3. `arc source update` shows "metafactory: failed"
4. `arc search` returns no results from default source

For any **new** `arc` install between F-1 landing and F-3 (Registry API client), the default source is effectively dead. Existing users are unaffected (their sources.yaml is already saved).

**Options:**
- (a) Add a guard in `fetchRemoteRegistry`: skip sources where `type === "metafactory"` with a log message like "metafactory source support coming in a future release"
- (b) Keep the old REGISTRY.yaml as a second default alongside the new metafactory source
- (c) Accept the gap — it's a dev-facing tool and F-3 is presumably next

No strong opinion on which — just want it to be a conscious decision rather than a surprise.

---

### Minor observations (non-blocking)

- **Double validation:** CLI calls `validateSource()`, then `addSource()` calls it again. Not harmful, just redundant. Could let `addSource` be the single validation gate and keep CLI lean.
- **Conditional spread** `...(opts.type !== "registry" ? { type: opts.type } : {})` — nice pattern, keeps sources.yaml clean for registry types.
- **Token plaintext on disk** — expected at this stage, just noting it for F-2 design consideration (whether to encrypt at rest or rely on filesystem permissions).

---

### Verdict

**Changes requested** — fix Finding 1 (registry URL validation regression), then this is good to merge. Finding 2 is a product decision, not a code defect.

Test coverage is thorough — 17 tests covering type parsing, URL rules, token security, backward compat, forward compat. Well-structured decomposition for the A-401 series.

jcfischer:
## Resolution Table

| # | Finding | Fix | Verification |
|---|---------|-----|-------------|
| 1 | Registry-type URL validation dropped (blocking) | Added URL scheme check for `registry` type in `validateSource()` -- rejects `garbage://`, `ftp://` etc. with same error message as before | 2 new tests: `rejects invalid URL scheme for registry type` + `rejects invalid URL scheme for implicit registry type` |
| 2 | Default source broken during F-1..F-3 window | Kept REGISTRY.yaml as second default (`metafactory-registry`). Added guard in `fetchRemoteRegistry` to skip `type:"metafactory"` sources gracefully (returns null) | Default now creates 2 sources. `arc search` works via fallback. Tests updated. |

**Double validation note:** Kept as-is for now -- `addSource` is the single validation gate for programmatic use, CLI pre-validates for better UX (error message before the throw). Can clean up later if it becomes friction.

**Full suite:** 320/320 passing.

@the-metafactory/luna -- ready for re-review.

mellanon (approved):
## Re-review: A-401 F-1 — Fix commit `98ac8ce2`

### Finding Resolution

| # | Finding | Status | Verification |
|---|---------|--------|-------------|
| 1 | Registry URL validation dropped | **Resolved** | `validateSource()` now checks `https://`, `http://`, `file://` for registry type. Two new tests cover explicit and implicit registry type rejection. Matches old behavior exactly. |
| 2 | Default source broken during F-1→F-3 | **Resolved** | Both (a) and (b): `fetchRemoteRegistry` skips `type:"metafactory"` with early `return null`, AND REGISTRY.yaml kept as fallback default (`metafactory-registry`). `arc search` continues to work for new installs. |

### Code verification

- Read `sources.ts` on branch — `validateSource()` has correct branching for both types, `getSourceType()` defaults properly, both default sources present
- Read `remote-registry.ts` on branch — type guard is at the top of `fetchRemoteRegistry`, correct placement before any network call
- Read `types.ts` — `SourceType` union and optional fields on `RegistrySource` are clean
- Diff: 320/320 tests, 19 new tests total across both commits

### One observation (non-blocking, cosmetic)

`arc source update` will show `metafactory: failed` for the API source since `updateAllSources` treats `null` returns as failures. Not incorrect (it _is_ non-functional until F-3), but a `skipped` status might be friendlier. Fine to address in F-3 when the metafactory client lands.

### Verdict

**Approved.** Both findings addressed cleanly, test coverage is thorough, transition path is well-handled. Good to merge.

PR #58, opened 2026-04-10
feat: add registry install path with SHA-256 verification (A-401 F-4)

jcfischer:
## Summary
- `arc install @scope/name[@version]` -- new registry install path
- Downloads tarball from metafactory storage endpoint, verifies SHA-256, extracts, installs
- Three-way CLI routing: package ref -> registry / URL -> git clone / simple name -> source lookup
- Backward compatible: existing git URL and simple name installs unchanged

## New library: src/lib/registry-install.ts
- `parsePackageRef` / `formatPackageRef` -- @scope/name[@version] handling
- `resolveFromRegistry` -- version + SHA-256 lookup via F-3 API client
- `downloadPackage` -- Bearer auth, 60s timeout, single retry
- `verifyChecksum` -- mandatory SHA-256, no bypass
- `extractPackage` -- tar extraction, manifest validation, cleanup on failure

## Security
- SHA-256 verification is mandatory for registry installs
- Checksum mismatch: abort + clear security warning
- Bearer tokens never logged
- Partial downloads cleaned up on failure

## Tests
13 new tests. Full suite: **365/365 passing**.

| Group | Tests | Covers |
|-------|-------|--------|
| parsePackageRef | 6 | Valid formats, null for URLs/paths/simple names |
| formatPackageRef | 2 | With/without version |
| verifyChecksum | 4 | Match, mismatch, case-insensitive, empty file |
| extractPackage | 1 | Invalid tarball rejection |

## SpecFlow
F-4 of A-401 series. Depends on F-1, F-2, F-3 (all merged).

F-5 (search across sources) is the final feature in this series.

@the-metafactory/luna -- requesting review

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -37,6 +37,14 @@ import {
 import type { CatalogEntry, ArtifactType, PackageTier, RegistrySource, SourceType } from "./types.js";
 import { login } from "./commands/login.js";
 import { logout } from "./commands/logout.js";
+import {
+  parsePackageRef,
+  formatPackageRef,
+  resolveFromRegistry,
+  downloadPackage,
+  verifyChecksum,
+  extractPackage,
+} from "./lib/registry-install.js";
 import { loadCatalog, saveCatalog, findEntry } from "./lib/catalog.js";
 import {
   loadSources,
@@ -74,17 +82,20 @@ program
     await ensureDirectories(paths);
     const db = openDatabase(paths.dbPath);
 
+    // Check if input is a registry package ref (@scope/name[@version])
+    const pkgRef = parsePackageRef(nameOrUrl);
+
     // Parse library:artifact colon syntax (only for non-URL names)
     const isUrl =
-      nameOrUrl.includes("/") ||
       nameOrUrl.startsWith("git@") ||
-      nameOrUrl.startsWith("http");
+      nameOrUrl.startsWith("http") ||
+      nameOrUrl.startsWith("file://");
 
     let libraryName: string | undefined;
     let artifactName: string | undefined;
     let lookupName = nameOrUrl;
 
-    if (!isUrl) {
+    if (!isUrl && !pkgRef) {
       const libRef = parseLibraryRef(nameOrUrl);
       if (libRef?.artifactName) {
         libraryName = libRef.libraryName;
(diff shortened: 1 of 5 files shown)


mellanon (changes requested):
## PR #58 Review — A-401 F-4: Registry Install with SHA-256 Verification

**Lenses applied:** CodeQuality, Security, Architecture, Performance

### Verdict: 🔴 Request Changes (1 blocking bug)

---

### 🔴 BLOCKING — `install()` integration will always fail for registry installs

**The problem:** After extracting the tarball, `cli.ts` calls `install()` with `repoUrl: formatPackageRef(...)` which produces `@metafactory/grove@1.0.0`. Inside `install()`:

1. `extractRepoName("@metafactory/grove@1.0.0")` → `"grove@1.0.0"`
2. `installPath = join(reposDir, "grove@1.0.0")` — **different** from the extracted dir `"metafactory__grove"`
3. `existsSync(installPath)` → `false`
4. Attempts `git clone @metafactory/grove@1.0.0 <path>` → **fails** (not a valid git URL)

The extracted package at `${reposDir}/metafactory__grove` is never found by `install()`.

**Suggested fix options (pick one):**
- **A)** Add an `installPath` option to `install()` that skips the clone step when a pre-extracted directory is provided
- **B)** Extract the post-clone steps (manifest read, symlinks, hooks, DB recording) into a shared function that both the git and registry paths can call
- **C)** Set `repoUrl` to the extracted directory path and ensure `extractRepoName` derives the same directory name

Option B is the cleanest architecturally — it separates "acquire source" from "process + install manifest".

---

### ⚠️ WARNING — URL construction without `encodeURIComponent`

In `resolveFromRegistry` (line 80):
```typescript
const versionDetailUrl = `${source.url}/api/v1/packages/@${ref.scope}/${ref.name}/versions`;
```

While the regex constrains scope/name to `[a-zA-Z0-9._-]` (no injection risk), this is inconsistent with F-3 which was specifically asked to add `encodeURIComponent` on path segments. Apply the same pattern here for consistency.

---

### ⚠️ WARNING — `isUrl` detection behavioral change

The original `isUrl` check included `nameOrUrl.includes("/")` which matched any URL with a slash. Removing it is necessary for `@scope/name` routing, but it also means bare-domain inputs like `github.com/user/repo` (without `https://`) would no longer route to git clone. Low risk since the documented usage is full URLs, but worth noting for edge-case awareness.

---

### 💡 SUGGESTION — Empty catch in `extractPackage`

`await unlink(tarballPath).catch(() => {})` silently swallows cleanup errors. Per repo rules, should at minimum use `.catch((_err) => {})` to signal intentional suppression.

### 💡 SUGGESTION — Missing test coverage for resolution + download

`resolveFromRegistry` and `downloadPackage` have no test coverage. These are the most integration-heavy functions. Even with mocked `fetch`, tests for version resolution fallback, retry behavior on 5xx, and auth-header attachment would catch regressions.

### 💡 SUGGESTION — Unbounded version capture

The regex `(?:@(.+))?$` accepts any string as version, including `@scope/name@1.0.0@evil`. Consider tightening to `(?:@([^@]+))?$` to prevent ambiguous parses.

---

### ✅ What's good

- **SHA-256 verification is mandatory** — no bypass, clean failure messaging, case-insensitive comparison. Security-first.
- **Three-way routing** — package ref / URL / name routing is well-structured and the parse-first approach is correct.
- **Error handling** — distinct messages for 401/403/404, single retry for transient failures, cleanup on checksum failure.
- **Type additions** (`PackageRef`, `VerifyResult`) are clean and well-placed in `types.ts`.
- **Test quality** — parsing, checksum, and extraction tests are thorough.

| # | Finding | Severity | Line |
|---|---------|----------|------|
| 1 | `install()` called with package ref will git-clone-fail | 🔴 Blocking | cli.ts:140 |
| 2 | URL path segments unencoded (inconsistent with F-3) | ⚠️ Warning | registry-install.ts:80 |
| 3 | `isUrl` behavior change for bare-domain inputs | ⚠️ Warning | cli.ts:87 |
| 4 | Empty catch should use `_err` naming | 💡 Suggestion | registry-install.ts:225 |
| 5 | No tests for `resolveFromRegistry` / `downloadPackage` | 💡 Suggestion | — |
| 6 | Unbounded version capture group | 💡 Suggestion | registry-install.ts:18 |

jcfischer:
## Resolution Table

| # | Finding | Fix |
|---|---------|-----|
| 1 | install() fails for registry installs (blocking) | Added `preExtractedPath` option to `InstallOptions`. When set, `install()` uses that path as installPath and skips the git clone step. CLI passes `extract.extractedPath` for registry installs. |
| 2 | URL encoding in resolveFromRegistry (warning) | `encodeURIComponent` applied to scope and name in version lookup URL |
| 3 | isUrl bare-domain behavior (warning) | Noted -- documented usage is full URLs. Low risk. |
| 4 | Empty catch in extractPackage (suggestion) | `catch((_err) => {})` with comment "cleanup failure is non-fatal" |
| 5 | No tests for resolveFromRegistry/downloadPackage (suggestion) | **10 new tests**: version resolution (latest, specific, not found, no metafactory source), download (success, Bearer auth, 401, 404, retry) |
| 6 | Unbounded version regex (suggestion) | Tightened to `(?:@([^@]+))?$`. Added test for `@scope/name@1.0.0@evil` rejection. |

Full suite: **375/375** passing.

@the-metafactory/luna -- ready for re-review.

mellanon (approved):
## Re-review: All findings resolved ✅

Fix commit `515835b` addresses all 6 items from the initial review:

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | **Blocking**: install() directory mismatch | `preExtractedPath` option bypasses git clone, passes extracted directory directly — clean minimal integration |
| 2 | URL encoding on scope/name | `encodeURIComponent` on both segments in version lookup URL |
| 3 | `isUrl` bare-domain behavior | Acknowledged — documented usage is full URLs; `@scope/name` routing happens first so no conflict |
| 4 | Empty catch on unlink | `_err` naming convention with non-fatal context |
| 5 | Missing test coverage for error paths | 10 new tests: resolveFromRegistry (latest, specific version, not found, no metafactory source) + downloadPackage (success, auth, 401, 404, retry) |
| 6 | Version regex allows multi-@ | Tightened to `(?:@([^@]+))?$` — test confirms `@scope/name@1.0.0@evil` → null |

**Architecture note:** The `preExtractedPath` approach is the right call — it's a minimal addition to `install()` that preserves all downstream logic (manifest reading, symlinks, hooks, DB registration) without duplicating it.

Suite: 375/375. No new issues. Ready to merge.

PR #64, opened 2026-04-13
feat: add version pinning for git-based installs (A-402)

jcfischer:
## Summary

- Parse `MySkill@1.2.0` name@version syntax in CLI input (`parseNameVersion`)
- Add `--version` flag for URL-based installs
- After git clone, checkout version tag (`v1.2.0` then `1.2.0` fallback)
- Helpful error with available tags when pinned version not found
- Registry installs (`@scope/name@1.2.0`) already worked — no changes needed

## Usage

```bash
arc install MySkill@1.2.0                              # name@version
arc install https://github.com/foo/bar --version 1.2.0 # URL + flag
arc install @scope/name@1.2.0                          # registry (existing)
```

## Changes

| File | Change |
|------|--------|
| `src/commands/install.ts` | `pinnedVersion` option, `parseNameVersion()`, `checkoutVersionTag()` |
| `src/cli.ts` | `--version` flag, wire name@version parsing to install |
| `test/commands/install.test.ts` | 8 new tests (3 git tag, 5 parseNameVersion) |
| `blueprint.yaml` | A-402 → done |
| `package.json` | 0.14.2 → 0.15.0 |

## Test plan

- [x] 467 tests passing (full suite)
- [x] Pinned install checks out correct v-prefixed tag
- [x] Pinned install checks out plain tag (no v prefix)
- [x] Missing tag fails with available versions list
- [x] parseNameVersion handles URLs, scoped refs, non-semver correctly
- [ ] Code review

### DIFF
--- a/blueprint.yaml
+++ b/blueprint.yaml
@@ -106,7 +106,7 @@ features:
 
   - id: A-402
     name: Version pinning
-    status: planned
+    status: done
     iteration: 4
     depends: [A-300]
     description: "arc install MySkill@1.2.0 — pin to specific version during install"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "arc",
-  "version": "0.14.2",
+  "version": "0.15.0",
   "description": "Agentic component package manager — install, manage, and distribute AI agent skills",
   "type": "module",
   "bin": {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -3,7 +3,7 @@
 import { Command } from "commander";
 import { createPaths, ensureDirectories } from "./lib/paths.js";
 import { openDatabase } from "./lib/db.js";
-import { install } from "./commands/install.js";
+import { install, parseNameVersion } from "./commands/install.js";
 import { list, formatList, formatListJson } from "./commands/list.js";
 import { info, formatInfo, formatInfoJson } from "./commands/info.js";
 import { audit, formatAudit } from "./commands/audit.js";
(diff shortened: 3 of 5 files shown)


mellanon:
## Review: feat: add version pinning for git-based installs (A-402)

**Verdict: Approve with minor notes**

Clean, focused feature. Well-tested. Good to merge.

### What's good
- **Correct scoping** — only touches git-based installs, leaves registry path alone
- **Smart tag fallback** — tries `v1.2.0` then `1.2.0`, helpful error listing available tags on miss
- **`parseNameVersion` is defensive** — excludes URLs, scoped refs, git@ prefixes, non-semver suffixes
- **CLI flag precedence** — `--version` flag overrides `@version` suffix via `??=`, correct
- **Tests are solid** — 8 tests covering tag checkout (v-prefix, plain, missing) and parse edge cases
- **Cleanup on failure** — `rm -rf` the cloned dir if version tag not found, no orphaned clones

1. **Inconsistent variable in URL path** — `cli.ts` line ~182 in the diff passes `pinnedVersion: opts.version` for URL-based installs, but the name-based path (line ~216) passes the local `pinnedVersion` variable. Not a bug (URLs won't trigger `parseNameVersion` so they'd be equal), but inconsistent — should use the local `pinnedVersion` in both paths for clarity.

2. **`--version` flag naming** — `arc install --version 1.2.0` could be confused with `arc --version` (show CLI version). Alternative names like `--pin` or `--tag` would be unambiguous. Low priority since Commander scopes flags per subcommand, but worth considering.

3. **No semver validation on `--version` flag** — `parseNameVersion` validates semver format, but the `--version` CLI flag passes straight to `git checkout` without validation. `arc install MySkill --version main` would checkout the `main` branch, not a version tag. Probably fine for power users, but adding the same semver check from `parseNameVersion` would make it consistent.

### Security
- `Bun.spawnSync` with argv array — no shell injection risk
- Version checkout scoped to the cloned repo directory — safe

mellanon (approved):
Approve — clean feature, well-tested. Minor notes in comment above.

jcfischer:
All three review notes addressed in 70eebf2:

1. **Consistent variable** — URL path now uses `pinnedVersion` local variable (same as name-based path)
2. **Flag renamed** — `--version` → `--pin` to avoid confusion with `arc --version`
3. **Semver validation** — `--pin` now validates with the same `/^v?\d+\.\d+/` check as `@version` suffix. `arc install MySkill --pin main` → error

467 tests still passing.

PR #67, opened 2026-04-13
fix: surface registry fetch failures instead of swallowing them (#50)

### LINKED ISSUE #50: Registry fetch failures are swallowed — user sees misleading 'not found in any source'
## Problem

When `fetchRemoteRegistry` fails to load a registry (non-OK HTTP, parse error, network failure), it silently returns `null` and moves on. The user-facing error then becomes:

```
❌ "grove" not found in any source. Try: arc search <keyword>
```

...which is misleading — the package *is* in the source, but arc couldn't read the source at all. The real failure (e.g. 401/404 from the Contents API) is never surfaced.

## Real-world impact

This bit us in `test-rig` CI: the `ECOSYSTEM_PAT` fine-grained token had clone access to `arc` but was missing repository access on `meta-factory`. The Contents API returned 404, `fetchRemoteRegistry` returned `null`, and the only signal was \`grove not found in any source\`. Debugging took a round-trip through local reproduction to figure out the token was the problem. The-metafactory/test-rig#(the recent e2e run) is the case study.

## Where

\`src/lib/remote-registry.ts\`, \`fetchRemoteRegistry\`:

- \`if (!response.ok) return null;\` — swallows HTTP errors with no hint of status code or URL
- \`catch {}\` block — swallows network errors, falls back to stale cache silently

## Proposed fix

When a fetch fails, log a warning to stderr with:
- source name
- URL attempted (rewritten, so the user sees the Contents API call)
- HTTP status + body snippet (for !ok) or error message (for network failures)
- hint that the token may lack repository access

Keep the \`return null\` behavior so commands still try other sources, but make the failure visible. Something like:

\`\`\`ts
if (!response.ok) {
  const body = await response.text().catch(() => "");
  console.warn(
    \`⚠  arc: failed to fetch source "\${source.name}" (\${response.status} \${response.statusText}): \${fetchUrl}\` +
    (body ? \`\\n   \${body.slice(0, 200)}\` : "") +
    (response.status === 401 || response.status === 404
      ? \`\\n   hint: check that GITHUB_TOKEN has 'Contents: Read' on the source repo\`
      : "")
  );
  return null;
}
\`\`\`

And similarly for the catch block.

## Follow-ups

- Consider an \`arc sources check\` command that explicitly probes each configured source and reports reachability + auth status. Would have turned this into a one-command diagnosis.

jcfischer:
## Summary
- Emit warnings to stderr when registry fetch fails (HTTP error, network failure, local file error)
- Include source name, URL, HTTP status, and response body snippet in warning
- Add auth hint for 401/404: "check that GITHUB_TOKEN has 'Contents: Read' on the source repo"
- Note stale cache fallback when network error + stale cache exists
- Keep `return null` behavior so commands still try remaining sources

## Before
```
❌ "grove" not found in any source. Try: arc search <keyword>
```

## After
```
Warning: failed to fetch source "metafactory-registry" (404 Not Found): https://api.github.com/repos/...
   {"message":"Not Found","documentation_url":"..."}
   hint: check that GITHUB_TOKEN has 'Contents: Read' on the source repo
❌ "grove" not found in any source. Try: arc search <keyword>
```

## Test plan
- [x] 476 tests pass (3 new tests added)
- [x] HTTP error test: verifies source name and status code in warning
- [x] Auth hint test: verifies 401 includes GITHUB_TOKEN hint
- [x] Network error + stale cache test: verifies fallback with warning

Closes #50

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "arc",
-  "version": "0.17.0",
+  "version": "0.17.1",
   "description": "Agentic component package manager — install, manage, and distribute AI agent skills",
   "type": "module",
   "bin": {
--- a/src/lib/remote-registry.ts
+++ b/src/lib/remote-registry.ts
@@ -64,7 +64,8 @@ export async function fetchRemoteRegistry(
       parsed.registry.components ??= [];
       parsed.registry.rules ??= [];
       return parsed;
-    } catch {
+    } catch (err: any) {
+      process.stderr.write(`Warning: failed to read local source "${source.name}": ${filePath}: ${err.message ?? err}\n`);
       return null;
     }
   }
@@ -112,7 +113,18 @@ export async function fetchRemoteRegistry(
       headers,
     });
 
-    if (!response.ok) return null;
+    if (!response.ok) {
+      const body = await response.text().catch(() => "");
+      const hint = (response.status === 401 || response.status === 404)
+        ? `\n   hint: check that GITHUB_TOKEN has 'Contents: Read' on the source repo`
+        : "";
+      process.stderr.write(
+        `Warning: failed to fetch source "${source.name}" (${response.status} ${response.statusText}): ${fetchUrl}` +
+        (body ? `\n   ${body.slice(0, 200)}` : "") +
+        hint + "\n",
+      );
+      return null;
+    }
 
     const text = await response.text();
     const parsed = YAML.parse(text) as RegistryConfig;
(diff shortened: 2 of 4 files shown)


mellanon:
## Review: fix: surface registry fetch failures instead of swallowing them (#50)

### Verdict: **Approve** ✅

Clean fix for a real UX problem — silent `return null` on registry fetch failures gave misleading "not found in any source" errors when the actual issue was auth or network. Now users see what actually went wrong.

### What's good

- **All four failure paths surfaced** — local file error, HTTP error, network failure, and stale cache fallback. Each gets a descriptive stderr warning with source name, URL/path, and error details.
- **Auth hint on 401 AND 404** — correct UX decision. GitHub returns 404 for private repos when tokens lack access (to avoid disclosing repo existence), so hinting about GITHUB_TOKEN permissions on both statuses is appropriate.
- **Response body truncation** — `body.slice(0, 200)` prevents massive error responses from flooding stderr. Good defensive measure.
- **Tests are well-structured** — each test properly captures/restores stderr and globalThis.fetch in `try/finally` blocks. Uses `forceRefresh: true` to bypass cache and exercise the actual fetch path.
- **Non-breaking** — all paths still `return null` after warning, preserving the existing fallback-to-next-source behavior.
- **PR description is excellent** — clear before/after examples, scoped test plan.

1. **Existing test will now emit stderr warnings** — the "returns null for unreachable URL" test (line 65) doesn't capture stderr, so it'll produce warning output during test runs. Not a bug, but test output will be noisier. Consider adding stderr capture there too for clean output, or not — it's a style preference.

2. **No test for the local file error path** — the `file://` branch (line 67 in diff) now warns on parse/read errors, but there's no test exercising that path. Low risk since it's a one-liner `stderr.write`, but noted for completeness.

3. **`catch (err: any)`** — `any` works but `unknown` with a narrowing pattern like `err instanceof Error ? err.message : String(err)` would be slightly more type-safe. The current `err.message ?? err` handles the common cases fine. Non-blocking.

### Security
- Response body in stderr (truncated to 200 chars) could theoretically include sensitive error detail, but this goes to local stderr only, not a public surface. Acceptable.
- No new attack surface.

PR #68, opened 2026-04-13
fix: fail loud on non-TTY stdin instead of silently half-installing (#46)

### LINKED ISSUE #46: arc install stalls on non-TTY stdin — should auto-default or fail loud
## Problem

`arc install <pkg>` prompts interactively for capability / hook consent even when stdin is `/dev/null` or otherwise non-interactive. With stdin ignored, the readline call returns EOF, arc silently treats that as "no", and the install completes as half-applied: package files land but hooks don't register.

## Repro (via test-rig Tier 0)

```bash
# test-rig/src/drivers/host.ts spawns bash with stdio: ["ignore", "pipe", "pipe"]
# test-rig/src/steps/install-grove.ts (before the workaround) ran:
bash -c 'export PATH="\$HOME/bin:\$PATH" && arc install grove'
```

Output captured during test-rig's first real end-to-end Tier 0 run:

\`\`\`
Found grove [component] in metafactory [community]

📦 Community source: metafactory

Install: Grove v0.21.2
Author: jcfischer (jcfischer)
Source: metafactory [community]
Risk: LOW

Capabilities:

📋 Grove wants to register hooks:
  • SessionStart → \${PAI_DIR}/hooks/GroveContext.hook.ts
  • PostToolUse → \${PAI_DIR}/hooks/EventLogger.hook.ts
  • Stop → \${PAI_DIR}/hooks/EventLogger.hook.ts
  • UserPromptSubmit → \${PAI_DIR}/hooks/EventLogger.hook.ts
  • PreToolUse (Bash) → \${PAI_DIR}/hooks/GroveBashGuard.hook.ts

Hooks run during Claude Code sessions.
Allow? [y/N]
\`\`\`

Then the process sat for ~18s before exiting. Exit code 0, but hooks were not actually registered.

## Current workaround

Pass `--yes` explicitly from test-rig:

\`\`\`ts
// test-rig/src/steps/install-grove.ts
\`arc install \${shellQuote(ctx.target)} --yes\`
\`\`\`

Works. But the default behaviour is unsafe in two ways:

1. **Silent half-install**: if an automated caller forgets `--yes`, arc silently pretends to succeed while leaving hooks unregistered. A downstream smoke test passes, the operator thinks the component is fully installed, but it isn't.
2. **Headless CI**: test-rig's Layer 1 / Layer 2 drivers (devcontainer, OrbStack) will hit the same thing — they'll all need to remember `--yes`. That's foot-gun-per-caller instead of foot-gun-per-arc.

## Expected

When `arc install` detects that `process.stdin.isTTY === false` (or Bun's equivalent), it should:

**Option A — fail loud**: print "arc install is interactive by default; pass --yes to run non-interactively" to stderr and exit non-zero. Forces the caller to be explicit.

**Option B — auto-default to safest-no-op**: treat non-TTY as refusal, print "non-TTY stdin detected; skipping capability confirmation and hook registration. Pass --yes to approve" on stderr, and exit non-zero so the caller knows it didn't fully install. Don't leave it ambiguous.

**Option C — match the UX of package managers**: `npm install` / `bun install` / `apt install -y` all treat non-TTY as "proceed with defaults". Arc would fail closed instead of open: non-TTY → same as \`--yes\` but with a stderr warning. This is the most ergonomic for CI but requires the caller to actively opt out.

I'd pick **Option A** — it's the safest and matches the "closed-by-default, publishing is a privilege" principle from the metafactory design decisions. Failing loud teaches callers to be explicit about auto-approval rather than hiding it in a default.

## Scope note

This is a test-rig follow-up from \`the-metafactory/test-rig#1\` Phase 0b validation. Not a blocker for Phase 0c / Layer 1 — test-rig's workaround works — but worth fixing in arc to remove the foot-gun.

jcfischer:
## Summary
- `arc install` now exits with a clear error when stdin is non-TTY and `--yes` not passed
- Previously: silently skipped hook registration, exited 0 — half-applied install
- Now: prints actionable error message and exits non-zero
- Defense-in-depth: `readLine()` also returns empty immediately on non-TTY
- Fixes typecheck errors in remote-registry test mock casts (pre-existing)

## Before (non-TTY)
```
Allow? [y/N]    ← hangs 18s, then exits 0 with hooks not registered
```

## After (non-TTY)
```
Error: arc install requires an interactive terminal for capability confirmation.
Pass --yes (-y) to approve non-interactively.
```

## Test plan
- [x] 477 tests pass (all existing tests use `yes: true`)
- [x] Existing install/library tests unaffected (they pass `--yes`)
- [ ] Manual: `echo | arc install grove` fails with clear error
- [ ] Manual: `arc install grove --yes` still works

Closes #46

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "arc",
-  "version": "0.17.1",
+  "version": "0.17.2",
   "description": "Agentic component package manager — install, manage, and distribute AI agent skills",
   "type": "module",
   "bin": {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -87,6 +87,13 @@ program
   .option("-y, --yes", "Skip confirmation prompt")
   .option("--pin <version>", "Pin to a specific version (git tag)")
   .action(async (nameOrUrl: string, opts: { yes?: boolean; pin?: string }) => {
+    // Non-TTY guard: fail loud rather than silently half-installing
+    if (!opts.yes && !process.stdin.isTTY) {
+      console.error("Error: arc install requires an interactive terminal for capability confirmation.");
+      console.error("Pass --yes (-y) to approve non-interactively.");
+      process.exit(1);
+    }
+
     const paths = createPaths();
     await ensureDirectories(paths);
     const db = openDatabase(paths.dbPath);
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -674,8 +674,12 @@ async function promptHookConsent(
 
 /**
  * Read a single line from stdin.
+ * Returns empty string immediately if stdin is not a TTY (defense in depth).
  */
 function readLine(): Promise<string> {
+  if (!process.stdin.isTTY) {
+    return Promise.resolve("");
+  }
   return new Promise((resolve) => {
     const stdin = process.stdin;
     stdin.setEncoding("utf-8");
--- a/test/unit/remote-registry.test.ts
+++ b/test/unit/remote-registry.test.ts
@@ -106,7 +106,7 @@ describe("fetchRemoteRegistry", () => {
 
   test("emits warning to stderr on HTTP error", async () => {
     const originalFetch = globalThis.fetch;
-    globalThis.fetch = (async () => new Response("Not Found", { status: 404, statusText: "Not Found" })) as typeof fetch;
+    globalThis.fetch = (async () => new Response("Not Found", { status: 404, statusText: "Not Found" })) as unknown as typeof fetch;
     (globalThis.fetch as any).preconnect = () => {};
 
     const stderrChunks: string[] = [];
@@ -134,7 +134,7 @@ describe("fetchRemoteRegistry", () => {
 
   test("emits warning with auth hint on 401", async () => {
     const originalFetch = globalThis.fetch;
-    globalThis.fetch = (async () => new Response("Unauthorized", { status: 401, statusText: "Unauthorized" })) as typeof fetch;
+    globalThis.fetch = (async () => new Response("Unauthorized", { status: 401, statusText: "Unauthorized" })) as unknown as typeof fetch;
     (globalThis.fetch as any).preconnect = () => {};
 
     const stderrChunks: string[] = [];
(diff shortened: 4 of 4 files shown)


mellanon:
## Review: Approve ✅

Clean, focused fix for a real foot-gun. Implements Option A from #46 exactly — fail loud, force callers to be explicit.

### What's good

- **Guard placement is correct** — checking at CLI entry (`cli.ts`) before any install work begins means no partial side effects (no directory creation, no DB open, no git clone) when the TTY check fails. The early-exit is the right pattern here.
- **Defense in depth** — the `readLine()` guard in `install.ts` is a good belt-and-suspenders measure. If a future code path somehow reaches `readLine()` without passing through the CLI guard, it returns empty instead of hanging. And it doesn't change behavior for the happy path.
- **Error message is actionable** — tells the user exactly what to do (`Pass --yes (-y)`). No ambiguity.
- **Test mock cleanup** — the `as any` pattern for mock fetch assignments is cleaner than the previous `as unknown as typeof fetch` / `.preconnect` two-step. Good opportunistic fix.

1. **`--pin` option appears in the diff's action signature** (`opts: { yes?: boolean; pin?: string }`) but doesn't appear on current main. If that's from a rebase onto another PR's changes, just confirming it's intentional and not an accidental inclusion.

2. **`process.exit(1)` vs throwing** — `process.exit(1)` works fine for a CLI tool, but if `arc install` is ever called programmatically (e.g., from test-rig importing the function directly), the exit would kill the parent process. Currently the guard is in `cli.ts` (the CLI entry point), not in the install function itself, so this is fine. Just noting that the install function in `install.ts` doesn't have its own TTY guard at the function level — the defense-in-depth is only in `readLine()`. This is the right layering for now.

3. **No test for the TTY guard itself** — understandable since testing `process.stdin.isTTY` and `process.exit` in a unit test is awkward. The manual test plan in the PR description covers it. The two manual checkboxes are unticked — worth confirming before merge.

### Security
- No new attack surface. The change is strictly additive (new guard) and the defense-in-depth path returns empty string (safe default).

**Verdict:** Approve — ready to merge after confirming the manual test checkboxes.

PR #70, opened 2026-04-13
feat: cosign binary bundling and wrapper (A-501, A-502)

jcfischer:
## Summary
- **A-501**: `scripts/fetch-cosign.ts` — downloads cosign v3.0.6 binaries for darwin-arm64, darwin-amd64, linux-amd64 from GitHub releases. Verifies SHA-256 against published checksums. Binaries stored in `vendor/cosign/` (gitignored).
- **A-502**: `src/lib/cosign.ts` — thin wrapper that detects platform, locates the bundled binary, and shells out to `cosign verify-blob --bundle` with certificate identity/OIDC issuer validation.
- Client-side foundation for Sigstore verification — A-503 will wire into `arc install` when meta-factory F5-500 lands.

## New files
| File | Purpose |
|------|---------|
| `scripts/fetch-cosign.ts` | Download + verify cosign binaries |
| `src/lib/cosign.ts` | Platform detection, binary resolution, `verifySigstoreBundle()` |
| `test/unit/cosign.test.ts` | Platform detection, binary resolution, error path tests |
| `vendor/cosign/.gitignore` | Exclude binaries from git (fetched on demand) |

## Design references
- DD-11: Sigstore/cosign for package artifacts
- DD-100: cosign keyless OIDC + Rekor v2 + Sigstore Bundle v0.3
- OQ-12: bundle strategy (embed binaries, no external dependency)

## Test plan
- [x] 482 tests pass (5 new cosign tests)
- [x] `detectPlatform()` returns valid os/arch/binaryName
- [x] `findCosignBinary()` returns string or null
- [x] `verifySigstoreBundle()` returns error when binary/artifact missing
- [ ] Manual: `bun scripts/fetch-cosign.ts` downloads and verifies all 3 binaries

Closes #69

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #71, opened 2026-04-13
feat: lazy auto-fetch cosign binary on first use

jcfischer:
## Summary
- `verifySigstoreBundle()` now auto-downloads cosign for the current platform if not present
- Downloads only the single binary needed (not all three platforms)
- SHA-256 verified against published checksums before writing
- `ensureCosignBinary()` exported for explicit pre-fetch
- `verifySigstoreBundle()` is now async (breaking change from #70)
- Download progress goes to stderr

## Before
```
cosign binary not found for cosign-darwin-arm64. Run: bun scripts/fetch-cosign.ts
```

## After
```
Downloading cosign v3.0.6 (cosign-darwin-arm64)...
cosign v3.0.6 downloaded and verified.
```

## Test plan
- [x] 485 tests pass (8 cosign tests)
- [x] `ensureCosignBinary()` returns path or error
- [x] `verifySigstoreBundle()` returns invalid for bad artifacts (with auto-fetch)
- [x] Platform detection includes download URL
- [ ] Manual: delete `vendor/cosign/cosign-*`, run verify — should auto-download

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "arc",
-  "version": "0.18.0",
+  "version": "0.19.0",
   "description": "Agentic component package manager — install, manage, and distribute AI agent skills",
   "type": "module",
   "bin": {
--- a/src/lib/cosign.ts
+++ b/src/lib/cosign.ts
@@ -108,14 +108,15 @@ export async function fetchCosignBinary(): Promise<{ path?: string; error?: stri
 
     const buffer = await response.arrayBuffer();
 
-    // Verify checksum
-    if (expectedHash) {
-      const hasher = new Bun.CryptoHasher("sha256");
-      hasher.update(buffer);
-      const actualHash = hasher.digest("hex");
-      if (actualHash !== expectedHash) {
-        return { error: `cosign checksum mismatch: expected ${expectedHash}, got ${actualHash}` };
-      }
+    // Verify checksum — hard fail if no checksum found
+    if (!expectedHash) {
+      return { error: `No checksum found for ${platform.binaryName} in cosign checksums — refusing to install unverified binary` };
+    }
+    const hasher = new Bun.CryptoHasher("sha256");
+    hasher.update(buffer);
+    const actualHash = hasher.digest("hex");
+    if (actualHash !== expectedHash) {
+      return { error: `cosign checksum mismatch: expected ${expectedHash}, got ${actualHash}` };
     }
 
     // Write binary
--- a/src/lib/cosign.ts
+++ b/src/lib/cosign.ts
@@ -97,6 +97,11 @@ export async function fetchCosignBinary(): Promise<{ path?: string; error?: stri
       }
     }
 
+    // Fail before downloading if no checksum available
+    if (!expectedHash) {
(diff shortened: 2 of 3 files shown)


mellanon:
## Review: feat: lazy auto-fetch cosign binary on first use

**PR:** #71 | **Author:** @jcfischer | **+147 / -27** | 2 files | 1 commit

---

### Verdict: **Approve with one security note**

Clean evolution of the cosign bundling from #70. Auto-fetch on first use is the right UX — no manual `bun scripts/fetch-cosign.ts` step. The async migration is safe since grep confirms no callers exist outside `cosign.ts` and its test file yet.

### What's good

- **Single-binary download** — only fetches the platform binary needed, not all three
- **SHA-256 verification** against published checksums before writing
- **Proper timeouts** — 30s for checksums, 120s for binary download
- **stderr messaging** — download progress doesn't pollute stdout
- **Clean error return pattern** — `{ path?, error? }` with no thrown exceptions
- **Tests properly async** with stderr suppression via `try/finally`
- **Breaking change documented** — PR body correctly flags the sync→async shift

### Notes

**1. Silent checksum skip (security — should fix)**

If the binary name isn't found in `cosign_checksums.txt`, `expectedHash` remains `undefined` and the checksum verification is silently skipped (lines ~107-113 in the diff). The binary gets written without any integrity check.

This should be a hard failure:

```ts
if (!expectedHash) {
  return { error: `No checksum found for ${platform.binaryName} in cosign checksums` };
}
```

Downloading an unverifiable binary defeats the purpose of the checksum fetch.

**2. Race condition (minor, non-blocking)**

Two concurrent calls to `ensureCosignBinary()` could both pass the `existsSync` check and race to download/write the same file. In practice both would write the same verified binary, and `writeFile` on the same path is effectively last-writer-wins, so this is benign. But if you wanted to be defensive, a simple lock (module-level `Promise` variable) would prevent the double download.

**3. `catch (err: any)` (non-blocking)**

Consistent with #70 pattern. Fine to leave.

### Summary

Solid implementation. The only actionable item is the missing checksum guard — currently a successful checksums fetch that doesn't contain the expected binary name results in an unverified download. Everything else is clean.

mellanon (approved):
## Re-review after commit `2cdc5c8`

**Previous note addressed:**

**1. Silent checksum skip — FIXED**
Hard failure now in place when `expectedHash` is `undefined`:
```ts
if (!expectedHash) {
  return { error: `No checksum found for ${platform.binaryName} in cosign checksums — refusing to install unverified binary` };
}
```
Exactly the right fix. No unverified binary can be written to disk.

**Minor ordering observation (non-blocking):** The check happens after the binary download, so bandwidth is wasted on a binary that will be rejected. Moving the `!expectedHash` check before the `fetch(platform.downloadUrl)` call would skip the unnecessary download. This is purely an efficiency nit — the security property (no unverified write) holds either way.

**2. Race condition — unchanged** (was non-blocking)
**3. `catch (err: any)` — unchanged** (was non-blocking)

### Verdict: **Approve**

Security concern resolved. The download-verify-write pipeline is now gapless.

PR #73, opened 2026-04-13
fix: refuse self-update in dev checkout (#72)

### LINKED ISSUE #72: arc self-update pulls into dev checkout — should detect install context
## Problem

`arc self-update` runs `git pull --ff-only` in the directory where arc's source lives (resolved from `import.meta.dir`). If the user installed arc via `bun link` from their development checkout, self-update modifies their working tree — potentially conflicting with in-progress work, uncommitted changes, or feature branches.

## Current behavior

```bash
# Developer working on arc in ~/work/mf/arc (on a feature branch)
arc self-update
# → git pull --ff-only in ~/work/mf/arc
# → Fails or overwrites dev work
```

## Expected behavior

`arc self-update` should detect the install context:

1. **Dedicated install** (e.g., `~/.config/metafactory/pkg/repos/arc`): `git pull` is correct — this is a managed installation.
2. **Dev checkout** (user's working tree with uncommitted changes or non-main branch): warn and refuse, or suggest the user update manually.

### Detection heuristic

- Check if CWD or `import.meta.dir` is inside `~/.config/metafactory/pkg/repos/` → safe to pull
- Otherwise check `git status --porcelain` for dirty state or `git branch --show-current` for non-main branch → warn and abort
- Or: check if the arc repo is a `bun link` target (symlink in `~/.bun/bin/`)

## Workaround

Developers manually `git pull` in their arc checkout.

## Context

Discovered during v0.19.0 session — ran `arc self-update` from the dev checkout and it showed "already up to date" (because the dev repo was already at HEAD), masking the fact that self-update was operating on the wrong directory.

jcfischer:
## Summary
- `arc self-update` now detects dev checkouts and refuses to pull
- Checks two signals: non-main branch, uncommitted changes
- Clear error messages directing the user to update manually

## Before (dev checkout on feature branch)
```
arc already up to date at v0.19.0 — no new commits on remote.
```
(Silently operates on dev checkout, masks the real state)

## After
```
Self-update failed: Refusing self-update: arc repo is on branch "feat/my-feature" (expected main).
This looks like a dev checkout — update manually with git.
```

## Test plan
- [x] 485 tests pass
- [ ] Manual: `arc self-update` on feature branch → refuses with branch name
- [ ] Manual: `arc self-update` with dirty tree → refuses with uncommitted changes message
- [ ] Manual: `arc self-update` on clean main → proceeds normally

Closes #72

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "arc",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "description": "Agentic component package manager — install, manage, and distribute AI agent skills",
   "type": "module",
   "bin": {
--- a/src/commands/self-update.ts
+++ b/src/commands/self-update.ts
@@ -216,7 +216,11 @@ function checkDevCheckout(pkgRoot: string): string | null {
     ? branchResult.stdout.toString().trim()
     : null;
 
-  if (branch && branch !== "main" && branch !== "master") {
+  if (!branch) {
+    return `Refusing self-update: arc repo is in detached HEAD state. This may indicate a tag-based checkout or interrupted rebase — update manually with git.`;
+  }
+
+  if (branch !== "main" && branch !== "master") {
     return `Refusing self-update: arc repo is on branch "${branch}" (expected main). This looks like a dev checkout — update manually with git.`;
   }
 
--- a/src/commands/self-update.ts
+++ b/src/commands/self-update.ts
@@ -10,6 +10,7 @@ export interface SelfUpdateResult {
 
 /**
  * Update arc itself by pulling latest from git and reinstalling deps.
+ * Refuses to update if running from a dev checkout (dirty tree or non-main branch).
  */
 export async function selfUpdate(): Promise<SelfUpdateResult> {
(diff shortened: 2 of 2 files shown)


mellanon (approved):
Clean, minimal guard — well placed and correct.

**What I checked:**
- Guard placement: after `pkgRoot`/`oldVersion` reads, before any `git pull` — correct
- `git branch --show-current` + `git status --porcelain` — standard, reliable signals
- Error messages are actionable and include context (branch name)
- `checkDevCheckout` is private (not exported) — good, it's an internal detail
- `checkSelfUpdate()` (read-only check) correctly has no guard — it only inspects, never modifies
- Return type fits existing `SelfUpdateResult` interface cleanly

**One edge case to be aware of (non-blocking):**
Detached HEAD (`git branch --show-current` returns empty string) — the branch check passes because `branch` is falsy. This means `arc self-update` would proceed on a detached HEAD with clean tree. Likely fine for tag-based installs, but worth knowing about. Could add `if (!branch)` as a third check if you want to be strict.

Ship it.

PR #75, opened 2026-04-14
feat(a-504): verify Ed25519 registry signature on install

jcfischer:
## Summary

Third client-side verification pass in the install pipeline — after SHA-256, before extract. Implements F-505 for the registry-level signing track (DD-12), consuming meta-factory's F-501 (shipped in #163) and the `manifest_canonical` field (shipped in #164).

## What

- **New:** `src/lib/registry-signing.ts`
  - `fetchRegistryPublicKey(source)` — hits `/.well-known/metafactory-signing-key`, returns null on 503/network-error/shape-mismatch.
  - `verifyRegistrySignature(pubB64, sigB64, bytes)` — Ed25519 verify via Web Crypto. Never throws, never uses `===` on bytes.
  - `verifyVersionSignature(source, signing, manifest_canonical)` — orchestrates fetch + key_id match + verify into a typed `{ verified, reason }` result.
- **Updated:** `src/lib/registry-install.ts`
  - `resolveFromRegistry` now calls `/packages/@scope/name@version` (returns sha256 + manifest_canonical + signing in one round-trip), replacing the older `/versions` list call.
  - `ResolvedRegistryPackage` carries `registrySignature`, `registryKeyId`, `manifestCanonical`.
- **Updated:** `src/cli.ts`
  - After SHA-256 passes, call `verifyVersionSignature`.
  - `verified === false` → exit 1 (no `--force` override, per DD-12/F-505).
  - `verified === null` → warn and proceed (legacy row or registry in degraded mode at publish time — mirrors meta-factory's graceful degradation).
- **Blueprint:** A-406 and A-504 marked `done`.

## Test plan

- [x] 17 new unit tests in `test/unit/registry-signing.test.ts`:
  - `verifyRegistrySignature`: round-trip, bit-flip (manifest), wrong key, malformed base64, wrong-length key (16 bytes), wrong-length sig (32 bytes).
  - `fetchRegistryPublicKey`: 200/503/network-error/shape-mismatch, anonymous (no Authorization header).
  - `verifyVersionSignature`: verified=true, verified=false on tamper, verified=null on unsigned, 503 while signed, key_id mismatch, missing manifest_canonical.
- [x] `registry-install.test.ts` mocks updated to the new per-version detail shape + one new test asserting signing fields propagate.
- [x] Full suite: 511/511 passing.
- [x] `bunx tsc --noEmit` clean.

## Dependencies satisfied

- meta-factory F-501 (PR #163) — registry signs, exposes signature
- meta-factory `manifest_canonical` (PR #164) — exposes exact signed bytes
- A-406 SHA-256 verify — already wired

## Not in this PR

A-503 (cosign/Sigstore verification) — separate track, depends on F-500 being wired end-to-end.

### DIFF
--- a/blueprint.yaml
+++ b/blueprint.yaml
@@ -135,10 +135,10 @@ features:
 
   - id: A-406
     name: SHA-256 registry verification
-    status: planned
+    status: done
     iteration: 5
     depends: [A-200]
-    description: "Verify sha256 field in REGISTRY.yaml entries against manifest content at install time"
+    description: "Verify sha256 field in registry version metadata against downloaded artifact at install time — implemented in src/lib/registry-install.ts verifyChecksum(), wired in src/cli.ts (non-bypassable)"
 
   - id: A-407
     name: Action artifact type
@@ -184,3 +184,10 @@ features:
     iteration: 6
     depends: [A-502, meta-factory:F5-500]
     description: "arc install fetches Sigstore bundle alongside artifact, calls cosign verify-blob --bundle --offline — blocks install on verification failure (DD-11, F-505)"
+
+  - id: A-504
+    name: arc install registry signature verification
+    status: done
+    iteration: 6
+    depends: [A-406, meta-factory:F-501]
+    description: "arc install fetches /.well-known/metafactory-signing-key from the registry and verifies the Ed25519 registry_signature bound to each version against manifest_canonical — blocks install on verification failure, no --force override. Unsigned/legacy versions emit a warning and proceed (matches meta-factory graceful degradation). Implemented in src/lib/registry-signing.ts + wired in src/cli.ts after SHA-256 check."
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -55,6 +55,7 @@ import {
   verifyChecksum,
   extractPackage,
 } from "./lib/registry-install.js";
+import { verifyVersionSignature } from "./lib/registry-signing.js";
 import { loadCatalog, saveCatalog, findEntry } from "./lib/catalog.js";
 import {
   loadSources,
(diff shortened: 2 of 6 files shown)


mellanon (approved):
## Review: A-504 — Ed25519 registry signature verification

**Verdict: Approved.** Solid supply-chain security work. The three-state result model (verified/rejected/not-applicable) is well-designed and the fail-closed posture for signed packages is the right call.

### What I checked

- `registry-signing.ts`: Key fetch (anonymous, no token leak to well-known), Ed25519 verify via Web Crypto (constant-time via `crypto.subtle.verify`, explicit key/sig length guards at 32/64 bytes), orchestration logic with all five early-return paths
- `registry-install.ts`: URL change from `/versions` list to `@version` per-version detail — response shape correctly maps the F-501 signing fields. Three new nullable fields on `ResolvedRegistryPackage` propagate cleanly
- `cli.ts`: Verification wired after SHA-256, before extract. Exit 1 on `verified === false` with no `--force` bypass. Temp file cleanup consistent with existing SHA-256 failure pattern
- `blueprint.yaml`: A-406 retroactively marked done (implementation was already in place), A-504 added with correct dependencies
- Tests: 17 new tests covering crypto round-trip, bit-flip rejection, wrong key, malformed base64, wrong-length key/sig, well-known 200/503/network/shape-mismatch, anonymous fetch, and all orchestration paths. Mock updates in registry-install.test.ts correctly reflect the new per-version detail shape

### One non-blocking note

The "This could indicate a compromised registry or a tampered manifest" message fires for **all** `verified === false` cases, including when the registry is simply unreachable (well-known returns 503/network error). That's more likely an infrastructure blip than a compromise. Consider distinguishing the two in future — e.g., "Registry signature could not be verified — public key endpoint unreachable" vs the scary tamper message for actual crypto failures. The `sigResult.reason` already carries the right detail, so the cli.ts message could branch on it. Not blocking — the behavior (fail-closed) is correct.

jcfischer:
Addressed in 511f035.  now branches on `sigResult.reason`: "public key unavailable" or "manifest_canonical missing" → "registry may be temporarily unreachable or misconfigured. Try again later." Everything else (signature mismatch, key_id mismatch) keeps the tamper warning. Fail-closed behavior unchanged.

PR #81, opened 2026-04-24
fix(bundle): expand defaults, warn on include trap, monorepo docs (#78)

### LINKED ISSUE #78: arc bundle: missing defaults + include-is-not-allowlist trap cause 278MB bundles in monorepos
## Symptoms

Running `arc bundle` on a monorepo root produced a **278MB tarball** (arc correctly rejected at the 50MB limit). After migrating to the library pattern (`type: library` at root, per-artifact `arc-manifest.yaml` in each package), `arc bundle packages/specflow` still produced **252MB**. Final size after adding explicit `bundle.exclude` for bun's compile cache: **197KB** (99.93% smaller).

Three distinct issues contributed. Each is small; together they make `arc bundle` feel broken on any non-trivial monorepo.

## 1. `DEFAULT_EXCLUSIONS` is missing common build/tool caches

Current defaults (`src/lib/bundle.ts:12`):

``````ts
export const DEFAULT_EXCLUSIONS = [
  ".git",
  "node_modules",
  ".env",
  ".env.*",
  "*.db",
  "*.sqlite",
  ".DS_Store",
  "Thumbs.db",
  ".specify",
  "test",
  "tests",
  "dist",
  "*.log",
  ".wrangler",
  ".claude",
];
``````

Missing patterns that are common enough to warrant defaults:

- `.*.bun-build` — bun's compiled-binary cache; 12 files × ~60MB each is typical for an actively-compiled project (this was the 689MB culprit in my case)
- `build`, `out` — besides `dist`
- `coverage`, `.nyc_output` — JS/TS test coverage
- `target` — Rust
- `.venv`, `__pycache__`, `*.pyc` — Python
- `*.tar.gz`, `*.tgz` — previous bundle artefacts left in the tree
- `.next`, `.turbo`, `.parcel-cache` — framework caches
- `.pnpm-store` — pnpm content-addressable store

**Proposed:** expand `DEFAULT_EXCLUSIONS` to cover these. Every one of them is something nobody would ever want to ship, and leaving them out makes \`arc bundle\` surprise users who don't read the source.

## 2. `bundle.include` is not an allowlist — but it looks like one

Current semantics (`src/lib/bundle.ts:60-67`, `src/lib/bundle.ts:118-126`):

``````ts
function buildTarArgs(exclusions: string[], includePatterns: string[]): string[] {
  const args: string[] = [];
  for (const pattern of exclusions) {
    if (!includePatterns.includes(pattern)) {  // ← only suppresses defaults
      args.push("--exclude", pattern);
    }
  }
  return args;
}
``````

`include` doesn't allow-list files into the tarball; it only cancels a matching default exclusion. The name is misleading. I spent real time writing an \`include: [packages/specflow/**]\` list thinking it would filter the bundle down. It didn't — \`include\` strings must match an entry already in \`DEFAULT_EXCLUSIONS\` verbatim to have any effect.

**Proposed:** one of

- **(a) Rename** the current field to something like \`allowDefaults\` or \`unexclude\`, and add a real \`include\` that acts as a tar-level include pattern (everything else excluded). This is the least-surprise fix.
- **(b) Document** the current semantics prominently in \`README\` and in the manifest schema's JSDoc. At minimum add a warning when \`bundle.include\` contains entries that don't match any default exclusion.

## 3. Monorepo / library pattern works but is undocumented

After digging through \`src/lib/manifest.ts\` and \`src/commands/install.ts\`, I discovered arc already has exactly the right abstraction for this: \`type: library\` at the repo root with an \`artifacts:\` list, plus per-artifact \`arc-manifest.yaml\`. With that pattern, \`arc bundle packages/specflow\` bundles only the specflow tree, not the whole monorepo.

But I only found it by reading the code. The README doesn't mention the library pattern, \`arc bundle --help\` doesn't mention it, and the bundle error message (\`Package tarball exceeds 50MB limit\`) gives no hint that the fix might be to add a path argument or restructure the manifest.

**Proposed:** 
- Add a short "Bundling a monorepo" section to the README
- Update \`arc bundle --help\` to show the \`[path]\` argument with a one-line example for monorepos
- When bundle rejects for size, suggest in the error: "If this is a monorepo root, try \`arc bundle <package-path>\` or convert to the library pattern."

## Repro

``````bash
git clone https://github.com/jcfischer/specflow-bundle
cd specflow-bundle
# before my fix:
#   git checkout <sha-before-ad44f58>
#   arc bundle          # 278MB, rejected
# after:
arc bundle packages/specflow    # 197KB, 144 files
``````

The fix commit: https://github.com/jcfischer/specflow-bundle/commit/ad44f58

## Suggested labels

\`bug\`, \`docs\`, \`dx\`

### PR DESCRIPTION
**Pull request by jcfischer**
Closes #78

## Summary

Three related DX issues caused `arc bundle` to produce enormous tarballs on monorepo roots (278MB) and monorepo package subdirs (252MB), before any user misconfiguration. Fixes all three.

### 1. Expanded `DEFAULT_EXCLUSIONS`

Added common build/cache patterns that nobody would intentionally ship:

- **JS / TS:** `build`, `out`, `coverage`, `.nyc_output`, `.next`, `.turbo`, `.parcel-cache`, `.pnpm-store`
- **Bun:** `.*.bun-build` (compiled-binary cache — the 689MB culprit in the filed issue)
- **Rust:** `target`
- **Python:** `.venv`, `__pycache__`, `*.pyc`
- **Prior bundles:** `*.tar.gz`, `*.tgz`

### 2. Warning for `bundle.include` trap

`bundle.include` looks like an allowlist but only **cancels** a matching default exclusion. Leaving the behavior intact for backward compat (specflow users rely on it) — now `createBundle` warns when `include` entries match no default, explaining the real semantics and pointing at `bundle.exclude` or a subdirectory path for filtering.

### 3. Monorepo / library pattern docs

The library pattern (`type: library` + `artifacts:` + per-artifact manifests) was fully implemented but undocumented. This PR:

- Adds a **Bundling a Monorepo** section to README with both patterns (subdir bundling + library root)
- Updates `arc bundle --help` to mention monorepo path + library pattern
- Enriches the size-rejection error with a concrete hint toward the library pattern and `bundle.exclude`

### Side fix: cosign test timeouts

Separate commit extends the 5s Bun default timeout on two cosign tests that do network I/O (Rekor/Fulcio verification, GitHub Release download). They were flaky on slow networks and in fresh CI/worktree environments. Pure test-harness change.

## Test plan

- [x] `bun test` — 577 pass, 0 fail
- [x] `bun src/cli.ts bundle --help` shows updated description
- [x] New unit tests:
  - `DEFAULT_EXCLUSIONS includes {pattern}` for each new default (30 assertions)
  - `excludes new build/cache defaults` — end-to-end tar output verification
  - `warns when bundle.include entries match no default exclusion`
  - `no orphan-include warning when bundle.include is empty or absent`
  - `warns with mix of matching and orphan includes`
  - `bundle.include overrides exclusions` — asserts no warning on matching include (backward compat)

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,18 +415,62 @@ The publish namespace is resolved in priority order:
 
 By default, these patterns are excluded from the tarball:
 
-`.git`, `node_modules`, `.env`, `.env.*`, `*.db`, `*.sqlite`, `.DS_Store`, `Thumbs.db`, `.specify`, `test`, `tests`, `dist`, `*.log`, `.wrangler`, `.claude`
+- **VCS / OS:** `.git`, `.DS_Store`, `Thumbs.db`
+- **Secrets:** `.env`, `.env.*`
+- **Databases / logs:** `*.db`, `*.sqlite`, `*.log`
+- **JS / TS build + cache:** `node_modules`, `dist`, `build`, `out`, `coverage`, `.nyc_output`, `.next`, `.turbo`, `.parcel-cache`, `.pnpm-store`
+- **Bun:** `.*.bun-build` (compiled-binary cache)
+- **Rust:** `target`
+- **Python:** `.venv`, `__pycache__`, `*.pyc`
+- **Prior bundle artefacts:** `*.tar.gz`, `*.tgz`
+- **arc / Cloudflare / Claude local state:** `.specify`, `.wrangler`, `.claude`
+- **Tests:** `test`, `tests` (override via `bundle.include` if your package ships tests)
 
-Override in `arc-manifest.yaml`:
+Extend or override in `arc-manifest.yaml`:
 
 ```yaml
 bundle:
   exclude:
     - "*.tmp"
-    - "coverage"
+    - "fixtures/large"
   include:
-    - "test"           # Override default exclusion
+    - "test"           # Cancels the default "test" exclusion
 ```
+
+#### `bundle.include` is not an allowlist
+
+`bundle.include` only **cancels** a matching default exclusion — an entry must appear verbatim in the default list above to have any effect. It does not filter the tarball down to a subset of files. Use `bundle.exclude` for that, or bundle a subdirectory directly (`arc bundle packages/my-pkg`). arc will warn when an `include` entry does not match any default.
+
+### Bundling a Monorepo
+
+If your repo is a monorepo with multiple publishable packages, there are two supported patterns:
+
+**1. Bundle one package at a time**
+
+Each package directory has its own `arc-manifest.yaml`. Run `arc bundle` against the subdirectory:
+
+```bash
+arc bundle packages/my-skill
+arc publish packages/my-skill
+```
+
+**2. Library root**
+
+Declare the repo root as a `library` with an `artifacts:` list. Each artifact is a subdirectory that contains its own `arc-manifest.yaml`.
+
+```yaml
+# /arc-manifest.yaml  (repo root)
+name: my-library
+version: 1.0.0
+type: library
+artifacts:
+  - path: packages/skill-a
+    description: Skill A
+  - path: packages/tool-b
+    description: Tool B
+```
+
+With the library pattern, `arc install` installs every artifact, and `arc bundle packages/skill-a` bundles only that subtree — ignoring the rest of the monorepo (including `node_modules`, build caches, sibling packages, etc.).
 
 ### Dry Run
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -813,7 +813,7 @@ program
 
 program
   .command("bundle [path]")
-  .description("Create a distributable tarball from a package directory")
+  .description("Create a distributable tarball from a package directory. For monorepos, pass a subdirectory (e.g. `arc bundle packages/my-pkg`) or use the library pattern at the repo root.")
   .option("-o, --output <path>", "Output path for the tarball")
   .action(async (path: string | undefined, opts: { output?: string }) => {
     const paths = createPaths();
--- a/src/lib/bundle.ts
+++ b/src/lib/bundle.ts
@@ -188,11 +188,13 @@ export async function createBundle(
   const exclusions = getExclusionPatterns(manifest);
   const includePatterns = manifest.bundle?.include ?? [];
 
-  const orphanIncludes = includePatterns.filter((p) => !DEFAULT_EXCLUSIONS.includes(p));
+  // Warn when include entries don't cancel any exclusion (default OR user-defined).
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Review by mellanon — APPROVED** [review]
## Re-review — APPROVE

Verified against issue #78 and `src/lib/bundle.ts`. All three sub-fixes land cleanly, scope matches the linked issue, and the test coverage is thorough.

### What I checked
- **`DEFAULT_EXCLUSIONS` expansion** (`src/lib/bundle.ts:12-46`) — every pattern from issue #78 is present, grouped with helpful comments. The `.*.bun-build` glob matches bun's compiled-cache pattern (the 689MB culprit).
- **Orphan-include warning** (`src/lib/bundle.ts:191-198`) — clear message, points users at the right fix (`bundle.exclude` or subdirectory bundling). Backward-compat preserved: matching includes (e.g. specflow's `test`) trigger no warning, asserted in `bundle.test.ts:337`.
- **Size-rejection hint** (`src/lib/bundle.ts:215-228`) — concrete and actionable, mentions both the library pattern and common cache offenders.
- **Docs** — README diff is well-organised (categorised exclusion list, separate "is not an allowlist" section, full monorepo section with both patterns). `arc bundle --help` description updated.
- **Cosign test timeouts** — sensible tactical fix; comments explain the network dependency. Long-term these probably belong behind an integration-test gate, but bumping to 60s unblocks the smoke-test gate without flakiness.
- **Tests** — 5 new `createBundle` tests cover: end-to-end exclusion of new defaults, orphan warning content, no-warning when include is absent, no-warning on matching include, mixed matching+orphan. The 30 `DEFAULT_EXCLUSIONS includes ${pattern}` cases are a clean regression guard.

### Non-blocking nit (file a follow-up if you agree)

The orphan-include detector compares against `DEFAULT_EXCLUSIONS` only:

```ts
// src/lib/bundle.ts:190
const orphanIncludes = includePatterns.filter((p) => !DEFAULT_EXCLUSIONS.includes(p));
```

But `buildTarArgs` (`bundle.ts:118-126`) lets `include` cancel **any** exclusion in the merged set — including user-defined `bundle.exclude` entries. So a manifest like:

```yaml
bundle:
  exclude: ["fixtures/large"]
  include: ["fixtures/large"]   # legitimately cancels the user's own exclude
```

…would emit a spurious "no effect" warning even though the include is functional. Tiny edge case, easy fix: `getExclusionPatterns(manifest).includes(p)` instead of `DEFAULT_EXCLUSIONS.includes(p)`. Not a blocker — happy to see this merged as-is and patched in a follow-up.

### Verdict
LGTM. 🟢 Ready to merge.

PR #82, opened 2026-04-24
fix(manifest): normalize network capability string shorthand (#79)

### LINKED ISSUE #79: toServerManifest silently produces {domain: undefined} when network capability uses string shorthand
## Summary

\`src/lib/publish.ts:75\` converts \`capabilities.network\` assuming every entry is \`{domain, reason}\`:

\`\`\`ts
const network = (caps.network ?? []).map((n) => ({ domain: n.domain }));
\`\`\`

\`ArcManifest.capabilities.network\` is typed as \`Array<{domain: string; reason: string}>\`. But the YAML parser does not enforce the shape — string entries pass through untouched:

\`\`\`yaml
capabilities:
  network:
    - github.com
    - agentskills.io
\`\`\`

arc accepts this, \`n\` is the string \`"github.com"\`, and \`n.domain\` is \`undefined\`. The server manifest ends up with \`network: [{domain: undefined}, {domain: undefined}]\` — stored as \`[{},{}]\` in D1 \`package_versions.manifest\`.

## Observed impact

Downstream on metafactory dev, the validation pipeline's L1 emit fed the empty objects into the HMAC audit-log payload. \`canonicalJson\` rejects \`undefined\`:

\`\`\`
[#197-B] Validation pipeline error for <submission>: TypeError: canonicalJson: undefined is not allowed
\`\`\`

Submission stuck in \`validating\` forever because metafactory's own \`.catch(console.error)\` swallows the throw (tracked as metafactory #303). From the publisher's perspective: \`arc publish\` reports success, package never becomes visible, no feedback.

## Reproduction

Fresh manifest with string-form network:
\`\`\`yaml
schema: arc/v1
name: test
version: 0.1.0
type: skill
namespace: jcfischer
license: MIT
capabilities:
  network:
    - github.com
\`\`\`

\`arc publish -s dev\` — succeeds, submission then hangs.

## Fix

Handle both shapes in \`toServerManifest\`:

\`\`\`ts
const network = (caps.network ?? []).map((n) =>
  typeof n === 'string'
    ? { domain: n }
    : { domain: n.domain }
);
\`\`\`

Additionally: tighten the \`ArcManifest\` zod/type to accept \`string | {domain, reason}\` for \`network\` (and mirror for other capability categories if similar shorthand is documented), or strict-parse and reject string form with a useful error.

## Related

- metafactory #303: validation pipeline silent failure swallowing errors like this one
- metafactory (to be filed): server accepts \`{domain: undefined}\` at publish time with no validation

jcfischer:
Closes #79

## Summary

`capabilities.network` YAML accepts string shorthand (`- github.com`) silently, even though `ArcManifest` types entries as `{domain, reason}`. All consumer sites deref `n.domain` → `undefined`. In `toServerManifest`, that sends `{domain: undefined}` to the registry, which D1 stores as `{}` and metafactory's validation pipeline rejects with `TypeError: canonicalJson: undefined is not allowed` — the submission hangs in `validating` forever while arc happily reports success.

## Fix

Normalize at the manifest boundary:

- **`readManifest`** runs `normalizeCapabilities`, which rewrites every string entry to `{domain, reason: ""}`, rejects malformed entries (non-string domain, null, number) with a useful error, and writes a stderr warning naming each shorthand entry so publishers can add a reason in the manifest.
- **`toServerManifest`** coerces defensively as well, so in-memory manifests that bypass `readManifest` (tests, programmatic callers) can never emit `{domain: undefined}`.

Downstream code (`formatCapabilities`, `db.ts`, `upgrade.ts`, `assessRisk`) keeps its typed assumption — no changes needed.

No behaviour change for manifests that already use object form.

## Test plan

- [x] `bun test` — 597 pass, 0 fail
- [x] `normalizeNetworkEntry` — 7 cases (string, object with/without reason, missing domain, number, null)
- [x] `normalizeCapabilities` — string rewrite, mixed shapes, no-op (no caps, no network, empty), invalid-entry throw, stderr warning emitted with correct content, no warning on pure object form
- [x] `readManifest` — string-form YAML parses to object form end-to-end
- [x] `toServerManifest` — object/string/mixed/undefined/null/number inputs all produce valid `{domain: string}` server entries; never emits `undefined`
- [x] Existing `manifest.test.ts` (object-form network) still passes
- [x] Existing `publish.test.ts` flow still passes

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -1,7 +1,7 @@
 # arc-manifest.yaml — capability declaration
 schema: arc/v1
 name: arc
-version: 0.21.0
+version: 0.21.1
 type: tool
 tier: core
 description: Agentic component package manager — install, manage, and distribute AI agent skills
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "arc",
-  "version": "0.21.0",
+  "version": "0.21.1",
   "description": "Agentic component package manager — install, manage, and distribute AI agent skills",
   "type": "module",
   "bin": {
--- a/src/lib/manifest.ts
+++ b/src/lib/manifest.ts
@@ -46,6 +46,8 @@ export async function readManifest(
         );
       }
 
+      normalizeCapabilities(parsed, filename);
+
       return parsed;
     } catch (err: any) {
       if (err.code === "ENOENT") continue;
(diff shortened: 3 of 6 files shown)


mellanon (approved):
## Verdict: APPROVE ✅

Clean, surgical fix for #79. Normalization at both boundaries (`readManifest` for YAML-loaded manifests, `toServerManifest` for in-memory/programmatic callers) is the right layering — downstream code (`formatCapabilities`, `db.ts`, `upgrade.ts`, `assessRisk`) keeps its typed invariant without any changes needed.

### Verified

- `normalizeNetworkEntry` covers string / object-with-reason / object-sans-reason / non-string-reason / no-domain / number / null — 7 cases, matches the helper's logic exactly
- `normalizeCapabilities` throws with a useful message on invalid entries, and emits a stderr warning naming each shorthand entry — publishers get actionable feedback
- `toServerManifest` uses `flatMap` so invalid entries are silently dropped rather than producing `{domain: undefined}` — the root cause from the issue. The regression test "never produces `{domain: undefined}` under any input shape" asserts this across 6 mixed inputs
- `readManifest` normalization happens after library-type early return, so the "libraries must not have capabilities" check still fires correctly
- Empty `network: []` correctly produces no `network` entry in server caps (existing behavior preserved)

### Nits (non-blocking)

1. **Silent drop asymmetry in `toServerManifest`**: the readManifest path *throws* on invalid entries, the toServerManifest path *silently drops* them. Reasonable as "belt-and-suspenders for programmatic callers", but a stderr warning on drop would make drift between the two paths observable. Follow-up only.

2. **Warning example cites only `shorthand[0]`**: with multiple shorthand entries, the example form `{domain: "${shorthand[0]}", reason: "..."}` might read as "only that one needs a reason." Minor wording polish — the full shorthand list is listed above in the warning so it's clear in context.

3. **Empty-domain (`{domain: ""}`) passes through**: out of scope for #79 (which was about `undefined`), but a manifest with `network: [{domain: "", reason: ""}]` would still reach the server. Worth a server-side validation ticket rather than adding scope here.

4. **Warning fires on every `readManifest` call**: if the CLI reads the manifest multiple times in one invocation (publish → list → upgrade flows), the warning may repeat. Not a blocker — trivial to dedupe later if it becomes noisy.

Ready to merge.

Issue #86, opened 2026-04-24
caduceus 0.3.1 install leaves dangling hook refs in settings.json — files never placed at registered paths


mellanon:
## Summary

Installing `caduceus` 0.3.1 via arc registers two Claude Code hooks in `~/.claude/settings.json` pointing at `${PAI_DIR}/hooks/...` paths, but the referenced files are never copied/symlinked there — they stay inside the caduceus repo dir. Result: every Claude Code session emits stderr errors on the Stop hook, non-blocking but noisy.

Related: [companion arc-verify blind-spot issue](https://github.com/the-metafactory/arc/issues/85). Filing here because the install-time partial-state is what creates the bug arc-verify can't catch.

## Repro

1. `arc install caduceus` (direct from https://github.com/jcfischer/caduceus).
2. Confirm the installer completed successfully (`arc list` shows `✅ caduceus v0.3.1 [active] (community)`).
3. Start any Claude Code session. Every turn closes with:

```
Stop hook error: Failed with non-blocking status code: /bin/sh: /Users/<me>/.claude/hooks/handlers/SkillNudge.ts: No such file or directory
```

Same class of error would fire on `SkillNudgeInject.hook.ts` if Stop ran that too.

## Evidence

- `~/.claude/settings.json` (modified by the arc/caduceus installer — **uncommitted** in the `~/.claude` git repo, i.e. directly written in place):

  - Line ~383: `"command": "${PAI_DIR}/hooks/SkillNudgeInject.hook.ts"`
  - Line ~562: `"command": "${PAI_DIR}/hooks/handlers/SkillNudge.ts"`

- `$PAI_DIR` expands to `~/.claude`.
- Neither file exists at the registered expanded paths (`~/.claude/hooks/SkillNudgeInject.hook.ts`, `~/.claude/hooks/handlers/SkillNudge.ts`).
- The actual files ship inside the caduceus repo:
  - `~/.config/metafactory/pkg/repos/caduceus/hooks/handlers/SkillNudge.ts`
  - `~/.config/metafactory/pkg/repos/caduceus/hooks/SkillNudgeInject.hook.ts`
- `~/.claude/skills/skill-manage/` is also absent (documented as used by the `Skill()` tool in caduceus's Hermes Pattern).
- `arc verify caduceus` reports `All checks passed` — see companion issue #85 for that blind spot.

## Root cause (my read)

[...]

Issue #85, opened 2026-04-24
`arc verify` doesn't validate `settings.json` hook paths — integrity-check blind spot


mellanon:
## Summary

`arc verify <pkg>` reports **all checks passed** even when a package has left dangling `settings.json` hook references pointing at non-existent files. This is the exact class of broken install the command is meant to catch, so the blind spot is load-bearing.

## Repro (on an arc v0.21.1 system that had a partially-failed install)

1. Install a package whose `arc-manifest.yaml` registers Claude Code hooks (e.g. caduceus 0.3.1 — see [companion issue](link-pending) for the specific install gap).
2. Observe that `settings.json` now contains hook entries like:

```json
{ "type": "command", "command": "${PAI_DIR}/hooks/handlers/SkillNudge.ts" }
```

…but `${PAI_DIR}/hooks/handlers/SkillNudge.ts` does not exist — the file ships inside the package's repo dir (`~/.config/metafactory/pkg/repos/<pkg>/hooks/handlers/SkillNudge.ts`) and no copy/symlink step landed it.

3. Run `arc verify <pkg>`. Observed output:

```
Verify: caduceus

  ✅ Repo directory exists (…/metafactory/pkg/repos/caduceus)
  ✅ arc-manifest.yaml valid
  ✅ Skill symlink valid (~/.claude/skills/caduceus)
  ✅ Git repo clean

All checks passed.
```

4. Claude Code sessions now emit one stderr line per turn:

```
Stop hook error: Failed with non-blocking status code: /bin/sh: /Users/<me>/.claude/hooks/handlers/SkillNudge.ts: No such file or directory
```

Non-blocking, so it keeps firing indefinitely. `arc verify` never surfaces it.

## Why this is load-bearing

The covenant `settings.json` expresses is: *"this command path is runnable."* Any verify step that doesn't walk those paths and confirm the targets resolve is not verifying integrity — it's verifying the repo checkout. A broken third-party installer (manifest + installer bug) can produce a state that passes verify but breaks every Claude Code turn, and the user has no tool to detect it short of reading error output.

## Proposed fix

Extend `arc verify <pkg>` to:

1. Read `~/.claude/settings.json` (the active config).
2. For every hook `command` that references `${PAI_DIR}/...` OR the package's own repo path OR a known packaging location, `stat` the target file.
[...]

PR #88, opened 2026-04-25
fix(install): honor provides.files across all types, gate hooks

### LINKED ISSUE #84: install silently drops provides.files for non-component types and wires hooks to missing files
## Summary

`arc install` silently drops `provides.files` for every artifact type except `component`, and silently registers `provides.hooks` whose command paths reference those (now-missing) targets. Result: install reports success, but the installed package is broken at runtime.

Companion to jcfischer/caduceus#1 — caduceus declared `type: skill` with `provides.files`, hit this behaviour, and every session now errors on its Stop hook.

## Reproduction

```bash
arc install https://github.com/jcfischer/caduceus -y
# ✅ Installed caduceus v0.3.1

# Check what the manifest promised vs. what got linked:
ls -la ~/.claude/hooks/handlers/SkillNudge.ts
#   ls: ...: No such file or directory
ls -la ~/.claude/skills/skill-manage/SKILL.md
#   ls: ...: No such file or directory

# But settings.json is happily pointing at the missing file:
jq '.hooks.Stop' ~/.claude/settings.json | grep SkillNudge
#   "command": "${PAI_DIR}/hooks/handlers/SkillNudge.ts"
```

Runtime, every session end:

```
Stop hook error: Failed with non-blocking status code:
/bin/sh: /Users/andreas/.claude/hooks/handlers/SkillNudge.ts: No such file or directory
```

## Root cause

`src/lib/artifact-installer.ts:61-193` — `provides.files` is only honoured in the `case "component"` branch (lines 107-117). The `case "skill"` branch and every other artifact type ignore the field:

```ts
case "component": {
  const files = manifest.provides?.files ?? [];
  for (const file of files) {
    const sourcePath = join(installDir, file.source);
    const targetPath = file.target.replace(/^~/, homedir());
    await mkdir(dirname(targetPath), { recursive: true });
    await createSymlink(sourcePath, targetPath);
  }
  break;
}

// skill / agent / prompt / tool / pipeline / action: no reference to provides.files
```

Meanwhile `provides.hooks` is wired up unconditionally via the hooks library and written into `settings.json`, pointing at the declared command paths — which for most packages are the same paths that `provides.files` was supposed to create.

Silent drop + unconditional hook registration = every cross-artifact package looks like it installs cleanly but fails at runtime.

## Why this matters

caduceus is the obvious example but the pattern will recur. "Multi-artifact" packages (a skill + a hook + a shared lib, or two skills in one repo) don't fit any single existing type cleanly:

- `skill` symlinks `skill/` subdir or repo root → can't place auxiliary files.
- `component` honours `provides.files` but doesn't surface CLI shims or skill triggers.
- `tool` only does bin shims.

So package authors reach for `provides.files` on whatever type seems closest to the "primary" artifact. Silently ignoring it punishes them at runtime with no diagnostic trail.

## Proposed fix (ranked)

**1. Validate at install time (minimal, safe).** Before returning success from the install path, for every `provides.files` entry verify:
- the source exists under `installDir`, and
- the target was actually created (symlink resolves to the source).

If any entry wasn't honoured, either error the install or print a loud warning listing the dropped entries. Separate check: every `provides.hooks[].command` should resolve to an existing file before being written into `settings.json`.

**2. Honour `provides.files` across all types.** Lift the file-symlink loop out of `case "component"` into a type-agnostic pre/post-pass so any package declaring files gets them. Then the per-type switch only handles the type's *own* primary layout (skill subdir, agent .md, pipeline dir, bin shims, etc.). Component's current body becomes "no primary layout, just the file pass."

**3. Long-term: formal multi-artifact type or `artifacts[]`.** Let a manifest declare an array of independent artifacts (skill-a, skill-b, hook-set, bin). The current single `type` is the bottleneck that pushes authors toward `provides.files` abuse. Out of scope for a quick fix.

Suggest doing (1) now to stop silent breakage, and (2) as the real fix in the same PR if possible.

## Tests to add

- `test/commands/install.test.ts` — install a manifest with `type: skill` and a `provides.files` entry; assert the symlink exists at the declared target, AND/OR assert install fails with a clear error.
- `test/commands/install.test.ts` — install a manifest whose `provides.hooks[].command` points at a non-existent path; assert install fails or warns, rather than silently wiring a broken hook.
- Regression: install the caduceus manifest fixture; assert all 8 declared targets resolve.

## Workaround for users hit today

```bash
# Remove the package (also cleans up the dangling hook registrations)
arc remove caduceus

# Install via the appropriate per-package workaround —
# for caduceus specifically, track jcfischer/caduceus#1 for a manifest fix
```

## Environment

- arc main @ `366e920` (2026-04-24)
- macOS, Bun, Claude Code

jcfischer:
## Summary

Closes #84. Refs #85, #86.

`arc install` previously honored `provides.files` only for `type: component` and silently dropped it for every other artifact type. Hook registration ran unconditionally, so packages declaring a `skill` (or `tool` / `agent` / …) plus `provides.files` plus `provides.hooks` looked like they installed cleanly but failed at runtime — `settings.json` pointed at files that were never placed. Caduceus is the canonical example.

## Changes

- **`src/lib/artifact-installer.ts`** — lift the `provides.files` symlink loop out of `case "component"` into a type-agnostic post-pass that runs after the per-type switch for every artifact type. Return `{ filesCreated, filesMissingSource }` so the caller can fail loudly when a declared source does not exist in the package.
- **`src/lib/hooks.ts`** — new `findMissingHookFiles()` helper that scans resolved hook commands (after `${PKG_DIR}` substitution) for absolute path tokens that don't exist on disk.
- **`src/commands/install.ts`** — wire both gates into the standalone install path and the library-artifact install path. Either gate failing returns `{ success: false, error: … }` with a precise diagnostic, never silently registering broken hooks.

## Tests

`test/commands/install.test.ts` (4 new):

- Skill-type package with `provides.files` → symlink created at target.
- Skill-type package with `provides.files` source missing → install fails with a clear error.
- Skill-type package with `provides.hooks` command pointing at a non-existent file → install fails, `settings.json` not mutated.
- Skill-type package with `provides.hooks` using `${PKG_DIR}` that resolves to an installed file → install succeeds and the hook is registered with the substituted path.

Full suite: 601/601 pass.

## Test plan

- [x] `bun test test/commands/install.test.ts` (25 pass)
- [x] `bun test` (601 pass)
- [ ] Manual smoke: `arc install <caduceus-shaped-manifest>` either succeeds with all targets in place or fails with the new diagnostic — tested via the new install tests against an isolated `createTestEnv()` repo.

### DIFF
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -277,7 +277,7 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
   );
   if (resolvedHooks?.length) {
     // Refuse to register hooks whose command points at a file that does not
-    // exist \u2014 silent registration of broken hooks was the original symptom of
+    // exist — silent registration of broken hooks was the original symptom of
     // issue #84.
     const missingHookFiles = findMissingHookFiles(resolvedHooks);
     if (missingHookFiles.length) {
@@ -543,7 +543,7 @@ export async function installSingleArtifact(
   );
   if (resolvedHooks?.length) {
     // Refuse to register hooks whose command points at a file that does not
-    // exist \u2014 silent registration of broken hooks was the original symptom of
+    // exist — silent registration of broken hooks was the original symptom of
     // issue #84.
     const missingHookFiles = findMissingHookFiles(resolvedHooks);
     if (missingHookFiles.length) {
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -5,7 +5,7 @@ import type { Database } from "bun:sqlite";
 import { readManifest, readLibraryArtifacts, assessRisk, formatCapabilities } from "../lib/manifest.js";
 import { recordInstall, getSkill } from "../lib/db.js";
 import { runScript } from "../lib/scripts.js";
-import { registerHooks, resolveHooksFromManifest } from "../lib/hooks.js";
+import { registerHooks, resolveHooksFromManifest, findMissingHookFiles } from "../lib/hooks.js";
 import { createArtifactSymlinks, resolveArtifactSourceDir, installNodeDependencies } from "../lib/artifact-installer.js";
 import { wireExtensions } from "../lib/extensions.js";
 import { extractRepoName } from "../lib/repo-name.js";
@@ -250,7 +250,7 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
   }
 
   // 4. Create symlinks based on artifact type
-  await createArtifactSymlinks({
+  const symlinkResult = await createArtifactSymlinks({
     type: manifest.type,
     manifest,
     paths,
@@ -258,6 +258,16 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
     consumerDir: opts.consumerDir,
     quiet: opts.yes,
   });
+  if (symlinkResult.filesMissingSource.length) {
+    const detail = symlinkResult.filesMissingSource
+      .map((f) => `  - ${f.source} -> ${f.target}`)
+      .join("\n");
+    return {
+      success: false,
+      error:
+        `Manifest declares provides.files entries whose source does not exist in the package:\n${detail}`,
+    };
+  }
 
   // 5b. Register hooks (if declared, with consent gating)
   const resolvedHooks = resolveHooksFromManifest(
@@ -266,6 +276,21 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
     manifest.name,
   );
   if (resolvedHooks?.length) {
+    // Refuse to register hooks whose command points at a file that does not
+    // exist \u2014 silent registration of broken hooks was the original symptom of
+    // issue #84.
+    const missingHookFiles = findMissingHookFiles(resolvedHooks);
+    if (missingHookFiles.length) {
+      const detail = missingHookFiles
+        .map((m) => `  - ${m.event}: ${m.command}\n      missing: ${m.missingPath}`)
+        .join("\n");
+      return {
+        success: false,
+        error:
+          `Manifest declares hooks whose command references a file that was not installed:\n${detail}\n` +
+          `Add the file to provides.files (or fix the command path) and reinstall.`,
+      };
+    }
     const tier = opts.sourceTier ?? manifest.tier ?? "custom";
     const approved = await promptHookConsent(
       manifest.name,
@@ -491,7 +516,7 @@ export async function installSingleArtifact(
   // Create symlinks based on artifact type
   const artifactType = manifest.type as ArtifactType;
 
-  await createArtifactSymlinks({
+  const symlinkResult = await createArtifactSymlinks({
     type: artifactType,
     manifest,
     paths,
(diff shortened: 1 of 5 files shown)


jcfischer:
Posting Luna's review on her behalf — she can't post directly.

> **Verdict: Approve** — solid, surgical fix. Matches the issue's recommended approach (validate at install + lift provides.files to all types). Diagnostics are precise, both install paths gated symmetrically, tests cover the four canonical scenarios.
>
> **What's good**
> - Type-agnostic post-pass is the right shape — per-type switch handles only the type's primary layout, `provides.files` runs uniformly. Component's case body collapsing to a no-op + comment makes intent obvious.
> - Both gates fail loud: `filesMissingSource` and `findMissingHookFiles` each return `{success:false}` with offending source/target/event listed. No more silent broken hooks.
> - `installSingleArtifact` mirrored — easy place to forget.
> - Hook resolution order correct: `findMissingHookFiles` runs after `${PKG_DIR}` substitution and after `createArtifactSymlinks`, so files just placed by `provides.files` can satisfy the hook check (caduceus shape works end-to-end).
> - 4th test pins down `${PKG_DIR}` substitution + `_pai_pkg` provenance.
>
> **Nits (non-blocking)**
> - Path-detection heuristic in `findMissingHookFiles`: `${HOME}` / `$HOME` not substituted upstream; shell redirects could false-positive (`cmd > /tmp/output.log`).
> - Partial state on failure: leftover symlinks when validation fails. Not a regression.
> - Inline comments at lines 261-263 and 522-524 contain literal `—` escape sequences rather than actual em-dashes.
>
> **Follow-ups (separate PRs)**
> - Rollback orphan symlinks on validation failure.
> - Wire `arc verify` (#85) to catch the same class of breakage post-install — defense in depth.

---

Addressed in commit c604bb8 (just pushed):

- ✅ **Em-dashes**: replaced literal `—` with actual `—` in `src/commands/install.ts` (both copies).
- ✅ **Shell redirects**: `findMissingHookFiles` now skips tokens that follow `>`, `>>`, `<`, `2>`, `2>>`, `&>`, `1>`, `|`. Added 5 unit tests in `test/unit/hooks.test.ts` (redirect skip, missing path, existing path, non-path tokens, quoted tokens).
- ✅ **`$HOME` TODO**: documented in the function header that env-var references aren't substituted upstream so a hook using `$HOME/script.sh` is left as a literal string and slips past inspection. Worth resolving when the resolver grows env-var substitution.
- 📝 **Orphan symlinks** → tracked separately as #89 (with two-pass validation vs. rollback options).
- 📝 **`arc verify` defense in depth** → already tracked as #85.

Full suite now 606/606 (was 601, +5 hook validation tests).

jcfischer:
Posting Luna's optional follow-ups for tracking:

> - Resolve `$HOME` / `${HOME}` upstream in `resolveHooksFromManifest` and drop the `~/` branch in `findMissingHookFiles` (paired change).
> - `arc verify` (#85) to catch this class post-install.

Filed as separate issues — both depend on #88 landing first:

- **#90** `hooks: resolve $HOME / ${HOME} upstream in resolveHooksFromManifest` — paired change with full repro, behavior-change note, and out-of-scope boundary (no arbitrary env-var allowlist).
- **#85** already tracks the `arc verify` defense-in-depth follow-up.

Not adding to this PR — Luna explicitly tagged them as "not for this PR" and the `$HOME` fix is a behavior change to `settings.json` content that deserves its own review.

PR #91, opened 2026-04-25
fix(install): send Bearer token to auth-gated storage endpoints

### LINKED ISSUE #83: arc install fails on auth-gated metafactory storage: downloadPackage never sends bearer token
## Summary

`arc install` cannot download from a `metafactory`-type source that requires auth on its storage endpoint (e.g. `dev.meta-factory.ai`). The tarball download path is hardcoded anonymous and never sends the bearer token from `arc login`, so even after successful login the install fails with `Access denied by storage endpoint.`

## Reproduction

```bash
# Configure an auth-gated metafactory source
arc source add metafactory-dev https://dev.meta-factory.ai --type metafactory --tier official
arc login --source metafactory-dev
# (token stored in sources.yaml, scope packages:read)

# Attempt to install a published package from that source
arc install @jcfischer/caduceus -y
```

Observed:

```
Found @jcfischer/caduceus v0.2.0 in metafactory-dev [official]
Downloading...
Access denied by storage endpoint.
```

Confirmed via direct probes:

| Request | Status |
|---|---|
| `GET /api/v1/packages/@jcfischer/caduceus@0.2.0` (anonymous) | 200 |
| `GET /api/v1/storage/download/<sha256>` (anonymous) | **401** `{"error":"Unauthorized","message":"Authentication required"}` |
| `GET /api/v1/storage/download/<sha256>` (with `Authorization: Bearer <token>`) | **200** (26514 bytes) |

So the hub is behaving correctly; arc isn't sending the bearer it already has.

## Root cause

`src/lib/registry-install.ts:143-155`:

```ts
/** Download a package tarball from the storage endpoint (anonymous) */
export async function downloadPackage(
  url: string,
  tempDir: string,
): Promise<DownloadResult> {
  ...
  const response = await fetch(url, {
    signal: AbortSignal.timeout(60_000),
  });
  if (response.status === 401 || response.status === 403) {
    return { success: false, error: "Access denied by storage endpoint." };
  }
```

No `Authorization` header is ever set, by design ("anonymous" in the doc comment). The `ResolvedDownload` returned by `resolveVersionFromSources` already carries the full `source: RegistrySource` (line 115), so the token is in hand at the call site — it just isn't plumbed through.

## Proposed fix (surgical)

Thread the source through so `downloadPackage` can attach `Authorization: Bearer <token>` when the source is `type: metafactory` and has a token:

- `src/lib/registry-install.ts`: change `downloadPackage(url, tempDir)` → `downloadPackage(url, tempDir, source)`; when `source.token && getSourceType(source) === "metafactory"`, set the header. Keep anonymous behaviour for `registry` sources and for sources without a token, so public-unauth packages still work.
- `src/cli.ts:163`: update the single caller to pass `resolved.source`.

Tests:
- Unit test for `downloadPackage` asserting header is sent when source has token, and omitted when it doesn't.
- Integration (createTestEnv + mock server): 401 anonymous / 200 authenticated path both covered.

## Workaround in the meantime

Install direct from the git URL — bypasses the hub storage entirely:

```bash
arc install https://github.com/<owner>/<repo>
```

Verified this works for the reproduction case above (caduceus v0.3.1 installed successfully).

## Related

- arc's `/api/v1/packages` listing endpoint works fine anonymously; only `/api/v1/storage/download/*` requires auth on dev. Prod (`meta-factory.ai`) may have the same requirement depending on deployment config — the fix is source-agnostic and should cover both.

jcfischer:
## Summary

Closes #83.

\`arc install\` could not download from a metafactory source whose \`/api/v1/storage/download/<sha256>\` endpoint requires auth (e.g. dev.meta-factory.ai). The download path was hardcoded anonymous and never attached the bearer token even after \`arc login\`, so install failed with \"Access denied by storage endpoint.\" The package detail endpoint already worked anonymously, so the token was sitting on the resolved source unused.

Verified by direct probes (per issue body): anonymous \`/api/v1/packages/...\` returns 200, anonymous \`/api/v1/storage/download/<sha256>\` returns 401, the same request with \`Authorization: Bearer <token>\` returns 200.

## Changes

- **\`src/lib/registry-install.ts\`** — extend \`downloadPackage(url, tempDir, source?)\` with an optional \`source\` param. When \`source.token\` is set and the source is metafactory-type, attach \`Authorization: Bearer <token>\`. Otherwise (no token, or registry-type source) the request stays anonymous so DD-80 public installs keep working.
- **\`src/cli.ts\`** — pass \`resolved.source\` through at the single call site.

Backwards-compatible: existing tests that call \`downloadPackage(url, tempDir)\` without a source argument keep their anonymous behaviour.

## Tests (test/unit/registry-install.test.ts, +4)

- Bearer token attached when source is metafactory with a token.
- Authorization header omitted when source has no token.
- Authorization header omitted for registry-type sources, even if they carry a token.
- 401 response still surfaces \"Access denied\" to the caller (e.g. expired/invalid token).

Full suite: 610/610 pass.

## Test plan

- [x] \`bun test test/unit/registry-install.test.ts\` (37 pass, 4 new)
- [x] \`bun test\` (610 pass)
- [x] \`bunx tsc --noEmit\` clean
- [ ] Manual smoke against dev.meta-factory.ai once merged: \`arc login --source metafactory-dev && arc install @jcfischer/caduceus -y\` should succeed where it currently fails with the access-denied error.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -158,9 +158,11 @@ program
 
       console.log(`Found ${formatPackageRef(pkgRef)} v${resolved.version} in ${resolved.source.name} [${resolved.source.tier}]`);
 
-      // Download (anonymous — no auth required per DD-80)
+      // Download. Anonymous by default (DD-80); the resolved source is passed
+      // through so an auth-gated metafactory storage endpoint receives the
+      // bearer token from `arc login` (issue #83).
       console.log(`Downloading...`);
-      const download = await downloadPackage(resolved.downloadUrl, paths.reposDir);
+      const download = await downloadPackage(resolved.downloadUrl, paths.reposDir, resolved.source);
       if (!download.success || !download.tempPath) {
         console.error(`${download.error}`);
         process.exit(1);
--- a/src/lib/registry-install.ts
+++ b/src/lib/registry-install.ts
@@ -140,17 +140,32 @@ export interface DownloadResult {
   error?: string;
 }
 
-/** Download a package tarball from the storage endpoint (anonymous) */
+/**
+ * Download a package tarball from the storage endpoint.
+ *
+ * When `source` is provided and is a metafactory source with a stored bearer
+ * token, attaches `Authorization: Bearer <token>` so the request can pass
+ * through an auth-gated storage endpoint (issue #83). For sources without a
+ * token, or registry-type sources, the request stays anonymous (DD-80) so
+ * public unauthenticated installs continue to work.
+ */
 export async function downloadPackage(
   url: string,
   tempDir: string,
+  source?: RegistrySource,
 ): Promise<DownloadResult> {
   const tempPath = join(tempDir, `arc-download-${Date.now()}.tar.gz`);
+
+  const headers: Record<string, string> = {};
+  if (source && source.token && getSourceType(source) === "metafactory") {
+    headers.Authorization = `Bearer ${source.token}`;
+  }
 
   let lastError: string | undefined;
   for (let attempt = 0; attempt < 2; attempt++) {
     try {
       const response = await fetch(url, {
+        headers,
         signal: AbortSignal.timeout(60_000),
       });
 
--- a/src/lib/registry-install.ts
+++ b/src/lib/registry-install.ts
@@ -140,6 +140,7 @@ export interface DownloadResult {
   error?: string;
 }
 
+/** Maximum number of HTTP redirects to follow on a single download. */
 const MAX_REDIRECTS = 5;
 
 /**
@@ -151,6 +152,10 @@ const MAX_REDIRECTS = 5;
  * WHATWG `fetch` spec already mandates this, but pinning the contract
  * here protects against runtime changes and makes the intent observable
  * in tests. Same-origin redirects keep the header.
+ *
+ * After MAX_REDIRECTS hops without reaching a non-3xx response, throw —
+ * never fall back to default fetch redirect-following, which would void
+ * the no-leak contract this function exists to enforce.
  */
 async function fetchFollowingRedirects(
   url: string,
@@ -158,7 +163,7 @@ async function fetchFollowingRedirects(
 ): Promise<Response> {
   let currentUrl = url;
   let currentHeaders = { ...init.headers };
-  for (let hop = 0; hop <= MAX_REDIRECTS; hop++) {
+  for (let hop = 0; hop < MAX_REDIRECTS; hop++) {
     const response = await fetch(currentUrl, {
       headers: currentHeaders,
       signal: init.signal,
@@ -180,10 +185,7 @@ async function fetchFollowingRedirects(
       currentHeaders = rest;
     }
   }
-  return fetch(currentUrl, {
-    headers: currentHeaders,
-    signal: init.signal,
-  });
+  throw new Error(`Too many redirects (>${MAX_REDIRECTS}) following ${url}`);
 }
 
 /**
--- a/src/lib/registry-install.ts
+++ b/src/lib/registry-install.ts
@@ -140,6 +140,52 @@ export interface DownloadResult {
   error?: string;
 }
 
+const MAX_REDIRECTS = 5;
+
+/**
+ * Fetch with explicit cross-origin Authorization stripping.
+ *
+ * If an auth'd request is redirected to a different origin (e.g. storage
+ * 302s to a presigned R2/S3 URL), drop the Authorization header on the
+ * next hop so the bearer token never reaches the redirect target. The
+ * WHATWG `fetch` spec already mandates this, but pinning the contract
+ * here protects against runtime changes and makes the intent observable
+ * in tests. Same-origin redirects keep the header.
+ */
+async function fetchFollowingRedirects(
+  url: string,
+  init: { headers: Record<string, string>; signal: AbortSignal },
+): Promise<Response> {
+  let currentUrl = url;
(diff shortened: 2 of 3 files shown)


jcfischer:
Forwarding Luna's review from Discord (she can't post directly without operator authorization).

---

**PR #91 — fix(install): send Bearer token to auth-gated storage endpoints**

One commit (`69d8406`), +116/-3, three files. Suite **610/610**, +4 unit tests. CI **MERGEABLE / CLEAN**.

**Verdict: Approve, ready to merge.** Surgical fix that matches the diagnosis in #83, preserves DD-80 anonymous installs, and covers all four behavioral branches with tests.

**What I verified**

1. ✅ **Diagnosis matches root cause.** Issue #83 documents anonymous probe = 401, Bearer probe = 200 against `/api/v1/storage/download/<sha256>`. Pre-fix `downloadPackage` at `src/lib/registry-install.ts:144-148` never set headers; the resolved source carries the token (`ResolvedRegistryPackage.source` at `:50`). Token plumbed through, no architectural shuffle.

2. ✅ **Type guard is correct.** `getSourceType(source) === "metafactory"` (`src/lib/sources.ts:61-63`) — defaults to `"registry"` when `source.type` is absent, so a misconfigured-as-registry source can never receive the bearer. Confirmed by the third test (`omits Authorization header for non-metafactory source even with token`) at `test/unit/registry-install.test.ts:471`.

3. ✅ **DD-80 preserved.** `source?` is optional; existing callers/tests at `:358, :377, :389, :402, :419, :518, :532` pass no source and stay anonymous. The `headers: Record<string, string>` starts empty; only populated when both `source.token` is truthy AND type is metafactory.

4. ✅ **Single call site updated.** Grep confirms one production caller (`src/cli.ts:163`); now passes `resolved.source`. No stale callers.

5. ✅ **Tests are well-targeted.** Four branches: token attached, no-token omits, registry-with-token omits, 401 surfaces "Access denied". Reuses the existing `metafactorySource()` helper (`test/unit/registry-install.test.ts:21`) consistent with the rest of the file.

6. ✅ **Empty-string defense.** `source.token` truthy check skips the header for `token: ""` — desired (an empty stored token shouldn't trigger auth).

**One thing worth flagging (non-blocking)**

🔒 **Bearer leakage on cross-origin redirect.** `fetch()` follows redirects by default. If `/api/v1/storage/download/<sha256>` ever 302s to a presigned R2/S3 URL on a different origin, the WHATWG/Bun runtime *should* strip the `Authorization` header on cross-origin redirect — but if that ever changes (or the implementation is run in a runtime that doesn't strip), the bearer leaks to R2. The current dev.meta-factory.ai path appears to stream directly (no redirect implied by the 200 in #83), so this is theoretical. Two cheap hardenings if you want belt-and-braces:

- Set `redirect: "manual"` on the auth'd fetch and re-issue the redirect target without the Authorization header, or
- Add a unit test asserting that on a 302 to a different origin, the second hop request has no `Authorization`. Bun's `fetch` honors the spec, so the test would just lock the contract.

Either is a follow-up, not a blocker for this PR.

**Optional follow-ups (not for this PR)**
- Same auth pattern is already in `fetchMetafactoryRegistry` / `fetchMetafactoryPackageDetail` — consider extracting a tiny `authHeaders(source)` helper (`src/lib/review.ts:6` already has one) so the three sites share a single rule. Pure refactor.
- Once #91 lands, the manual smoke against `dev.meta-factory.ai` in the test plan should be checked off in the PR description before close.

jcfischer:
Addressed Luna's non-blocking redirect-leak note in commit db52d85.

**What changed**
- New `fetchFollowingRedirects()` wrapper in `src/lib/registry-install.ts` walks redirects manually (`redirect: "manual"`, cap 5 hops). On each hop, drops the `Authorization` header iff the next URL is a different origin. Same-origin redirects keep the bearer (internal version-pinning redirects must still authenticate).
- `downloadPackage` now routes through it for both auth'd and anonymous downloads.

**Tests (+2)**
- 302 from `meta-factory.test` → `r2.cloudflarestorage.com`: first hop carries the bearer, second hop has no Authorization header.
- 307 same-origin redirect: bearer preserved across both hops.

Suite now **612/612**, tsc clean. Locks the WHATWG/Bun fetch contract so a runtime change can't silently leak the token.

Optional follow-ups (`authHeaders` helper + manual smoke checkbox) deferred to separate PR — pure refactor + post-merge action.

Pinging Luna for re-review.

jcfischer:
Addressed Luna's two nits in commit 8ce2252.

- ✅ **Loop bound**: changed `hop <= MAX_REDIRECTS` → `hop < MAX_REDIRECTS`. `MAX_REDIRECTS = 5` now means 5 hops, matching curl/undici defaults.
- ✅ **Fallback fetch contract**: removed the post-loop `return fetch(...)` that silently auto-followed past the cap. Now throws `Too many redirects` instead — never voids the no-leak contract.
- ➕ **Test (+1)**: server that 302s forever to fresh cross-origin hosts. Install fails, total fetch attempts stay within `cap × retries` (≤10). No request past the cap is ever auto-followed.

Suite: **613/613**, tsc clean.

Filed Luna's optional follow-up as **#92** — extract `fetchFollowingRedirects` to `src/lib/auth-fetch.ts` for reuse by `fetchMetafactoryRegistry` / `fetchMetafactoryPackageDetail` (same risk profile). Depends on this PR landing.

Pinging Luna for re-review.

Issue #89, opened 2026-04-25
install: rollback partial symlinks when provides.files validation fails


jcfischer:
## Summary

Follow-up from Luna's review of #88. When \`createArtifactSymlinks\` is partway through the type-agnostic \`provides.files\` pass and finds entry N's source missing, entries 1..N-1 already created symlinks and the per-type primary symlinks (skill dir, bin shim, agent .md, etc.) are also already in place.

\`install\` now correctly returns \`{success: false, error: ...}\` and refuses to record the package in the DB or register hooks. But the orphan symlinks under \`~/.claude/{skills,agents,commands,bin}/\` and any \`provides.files\` targets that were created remain on disk. The user has to clean them up manually or run \`arc remove\` (which won't find a DB entry to drive the cleanup).

## Repro

Manifest with two \`provides.files\`, second source intentionally missing:

\`\`\`yaml
provides:
  files:
    - source: src/exists.ts
      target: ~/.claude/handlers/exists.ts
    - source: src/missing.ts        # not in repo
      target: ~/.claude/handlers/missing.ts
\`\`\`

After failed install:
- \`~/.claude/skills/<name>\` symlink exists.
- \`~/.claude/handlers/exists.ts\` symlink exists.
- DB has no entry, hook gate did not run.

## Proposed fix

Two options:

1. **Two-pass validation** — first walk all \`provides.files\` and assert every source exists, only then start creating symlinks. Cheap, prevents the partial state class entirely.
2. **Rollback on failure** — \`createArtifactSymlinks\` tracks every symlink it creates (already returns \`filesCreated\`); install wraps the call in a try/finally that unlinks them on the error path along with the per-type primary symlinks.

Option 1 is simpler and addresses the common case (manifest typo, repo drift). Option 2 is more general but couples \`install\` more tightly to the per-type symlink details.

## Related

- Luna's review on #88
- #85 (arc verify blind spot — would catch this class post-hoc as defense in depth)

## Labels

bug, next

PR #93, opened 2026-04-25
fix(verify): walk settings.json hook paths, surface missing targets

### LINKED ISSUE #85: `arc verify` doesn't validate `settings.json` hook paths — integrity-check blind spot

jcfischer:
## Summary

Closes #85.

\`arc verify\` previously reported all-green even when a package left dangling \`settings.json\` hook references pointing at non-existent files (the exact symptom that drove caduceus 0.3.1's broken install). That blind spot was load-bearing: \`settings.json\` expresses a covenant — \"this command path is runnable\" — and verify wasn't walking those paths.

## Changes

**\`src/commands/verify.ts\`** — add a fifth check:

1. Enumerate the package's registered hooks via \`listPackageHooks(name, settingsPath)\`.
2. Run \`findMissingHookFiles()\` (introduced in #88) over them.
3. Pass when all command paths resolve. Fail when any don't, with a per-hook detail line: \`event: command\\n      missing: <path>\`.
4. **Hint logic**: when a missing target's basename can be located somewhere under the package's repo dir (depth ≤ 4, skip \`node_modules\`/dotdirs), attach a hint pointing at the repo-relative path plus a \"add a \`provides.files\` entry to land it at the hook target\" suggestion. Matches the recall/miner fix pattern referenced in the issue.

## Tests (test/commands/verify.test.ts, +3)

- Package with no hooks → verify passes, hook check omitted.
- Package with hooks whose targets resolve (via \`provides.files\` landing the file) → hook check passes.
- Package whose hook target was deleted post-install → verify **fails**, detail includes the missing path AND a hint pointing at the file's location in the repo dir AND the \`provides.files\` suggestion.

Full suite: 616/616 pass (+3 new).

## Test plan

- [x] \`bun test test/commands/verify.test.ts\` (6 pass, 3 new)
- [x] \`bun test\` (616 pass)
- [x] \`bunx tsc --noEmit\` clean
- [ ] Manual smoke once merged: re-install a hook-shaped package, mutate \`settings.json\` to point at a non-existent file, run \`arc verify <pkg>\`, expect a clear failure with the hint.

## Out of scope (mentioned in issue, deferred)

- Synthetic hook-event smoke test gated behind \`--hooks\` flag (issue's optional point 4). Useful for CI but a larger surface — separate PR if there's appetite.
- Wiring \`arc install\` to run extended verify as a post-install gate with rollback. Would close the partial-install loophole at source; depends on this PR + #89 (orphan-symlink rollback).

### DIFF
--- a/src/commands/enable.ts
+++ b/src/commands/enable.ts
@@ -102,11 +102,13 @@ export async function enable(
     }
   }
 
-  // Re-register hooks when enabling (consent was given at install time)
+  // Re-register hooks when enabling (consent was given at install time).
+  // paths.claudeRoot is threaded as $PAI_DIR expansion target — see install.ts.
   const resolvedHooks = resolveHooksFromManifest(
     manifest?.provides?.hooks,
     skill.install_path,
     name,
+    paths.claudeRoot,
   );
   if (resolvedHooks?.length) {
     const settingsPath = paths.settingsPath;
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -269,11 +269,15 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
     };
   }
 
-  // 5b. Register hooks (if declared, with consent gating)
+  // 5b. Register hooks (if declared, with consent gating).
+  // paths.claudeRoot is passed as the $PAI_DIR expansion target so a hook
+  // command like ${PAI_DIR}/hooks/handlers/Foo.ts gets stat'd against the
+  // absolute path the runtime would resolve to (issue #85).
   const resolvedHooks = resolveHooksFromManifest(
     manifest.provides?.hooks,
     installPath,
     manifest.name,
+    paths.claudeRoot,
   );
   if (resolvedHooks?.length) {
     // Refuse to register hooks whose command points at a file that does not
@@ -535,11 +539,13 @@ export async function installSingleArtifact(
     };
   }
 
-  // Register hooks (with consent gating — same as standalone install)
+  // Register hooks (with consent gating — same as standalone install).
+  // See note above on paths.claudeRoot threading $PAI_DIR substitution.
   const resolvedHooks = resolveHooksFromManifest(
     manifest.provides?.hooks,
     artifactDir,
     manifest.name,
+    paths.claudeRoot,
   );
   if (resolvedHooks?.length) {
     // Refuse to register hooks whose command points at a file that does not
--- a/src/commands/upgrade.ts
+++ b/src/commands/upgrade.ts
@@ -224,11 +224,13 @@ export async function upgradePackage(
     });
   }
 
-  // Re-register hooks (remove old, add new) — no consent prompt on upgrade
+  // Re-register hooks (remove old, add new) — no consent prompt on upgrade.
+  // paths.claudeRoot is threaded as $PAI_DIR expansion target — see install.ts.
   const resolvedHooks = resolveHooksFromManifest(
     manifest.provides?.hooks,
     installPath,
     name,
+    paths.claudeRoot,
   );
   if (resolvedHooks?.length) {
     const settingsPath = paths.settingsPath;
--- a/src/commands/verify.ts
+++ b/src/commands/verify.ts
@@ -1,5 +1,5 @@
 import { join, relative } from "path";
-import { existsSync } from "fs";
+import { existsSync, readdirSync, statSync } from "fs";
 import type { Database } from "bun:sqlite";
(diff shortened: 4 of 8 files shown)


jcfischer:
Addressed Luna's substantive nit + cosmetic in commit ee74953.

**Substantive — `${PAI_DIR}` slipping past:**

Fixed at the resolver level per Luna's preferred shape — closes both install-gating (#84) and verify-time detection (#85) in one place. `resolveHooksFromManifest` now accepts an optional `paiDir` and substitutes `$PAI_DIR` / `${PAI_DIR}` alongside the existing `$PKG_DIR` and `$<NAME>_DIR`. `install`, `enable`, and `upgrade` thread `paths.claudeRoot` through. settings.json now stores the absolute path the runtime would resolve to.

Two regression tests added in `test/commands/install.test.ts`:
- `${PAI_DIR}/handlers/SkillNudge.ts` with no `provides.files` to land it → install fails, diagnostic includes the resolved absolute path.
- Same shape with `provides.files` landing the file → install succeeds, settings.json contains the absolute path (no literal `${PAI_DIR}`).

**Cosmetic — `require("fs")` inside `walk`:**

Lifted to top-level ESM import. Bonus: `findFileInRepo` now short-circuits globally on first match (Luna's other cosmetic, addressed gratis).

**Other cosmetic — multiple matches:** kept as is; depth-4 bound makes the cost trivial and the global short-circuit means siblings stop walking.

**Tests:** +3 unit (`$PAI_DIR` substitution, back-compat, composition with other vars), +2 install regression. Suite **621/621**, tsc clean.

**Follow-up issue not filed** — this PR closes the substitution gap Luna pointed at, so #85's motivating shape is fully covered. No separate issue needed.

PR #94, opened 2026-04-25
fix(manifest): improve missing-capabilities error with example

### LINKED ISSUE #87: bundle: improve error message when 'capabilities' missing on non-component manifest
## Repro

```yaml
# arc-manifest.yaml
name: grove
namespace: the-metafactory
version: 0.23.0
type: tool
# ... no `capabilities:` block
```

```bash
$ arc bundle .
error: Invalid arc-manifest.yaml: missing required field 'capabilities' (only optional for type: component)
      at readManifest (src/lib/manifest.ts:44:19)
```

## Problem

The current message tells you *what* is missing but not *what to write*. A first-time publisher of a `tool` package has to grep the arc source / look at sibling repos to discover:

- `capabilities` is an object with `filesystem` and/or `network` keys
- `filesystem[]` entries are `{path, access, reason}` (access ∈ `read` / `read-write`)
- `network[]` entries are either bare domain strings or `{domain, reason}` objects
- `reason` is required for audit / install-time prompting
- `rules` type also exempt (per the source)

## Proposed

Show a minimal canonical block in the error:

```
error: Invalid arc-manifest.yaml: missing required field 'capabilities'
       Required for type: tool, agent, skill, library (optional only for: component, rules).

       Minimal example:

       capabilities:
         filesystem:
           - path: ~/.config/<your-tool>/
             access: read-write
             reason: <what is stored here>
         network:
           - domain: api.example.com
             reason: <what calls this domain>

       See: https://github.com/the-metafactory/arc/blob/main/README.md#capabilities
```

Or at minimum link directly to the schema doc and list the exempted types.

## Why

- 5+ minute fix (template literal in `src/lib/manifest.ts:44-46`) for every-time-someone-types-arc-bundle savings
- Reduces Discord pings asking 'what shape does capabilities want'
- Same pattern would help for the library-validation error a few lines down (`Libraries must have an artifacts array...`) — show the canonical shape there too

## Related

- Hit while publishing `@the-metafactory/grove@0.23.0` after switching from `type: component` to `type: tool`.

jcfischer:
## Summary

Closes #87.

When \`capabilities\` is missing on a non-component/non-rules manifest, the previous error told a first-time publisher *what* was missing but not *what to write*. They had to grep the source or copy from sibling repos to discover the schema.

## Changes

**\`src/lib/manifest.ts\`** — replace the one-line error with a multi-line message:

\`\`\`
Invalid arc-manifest.yaml: missing required field 'capabilities'
Required for type: skill, tool, agent, prompt, pipeline, action, system (optional only for: component, rules).

Minimal example:

capabilities:
  filesystem:
    read: []                  # paths the package reads from
    write: []                 # paths the package writes to
  network:
    - domain: api.example.com
      reason: <what calls this domain>
  bash:
    allowed: false            # set true if the package shells out
  secrets: []                 # env vars or secret keys the package reads

See: https://github.com/the-metafactory/arc/blob/main/README.md#capabilities
\`\`\`

Also corrects the prior message which said \"only optional for type: component\" but \`rules\` was already exempt in the code.

## Tests (test/unit/manifest.test.ts, +1)

Asserts the error contains: \`'capabilities'\`, both exempt types (\`component\`, \`rules\`), the canonical block subkeys (\`filesystem\`, \`network\`, \`bash\`, \`secrets\`, \`reason\`), and the README link.

Suite: 622/622 pass, tsc clean.

### DIFF
--- a/src/lib/manifest.ts
+++ b/src/lib/manifest.ts
@@ -44,7 +44,7 @@ export async function readManifest(
         throw new Error(
           [
             `Invalid ${filename}: missing required field 'capabilities'`,
-            `Required for type: skill, tool, agent, prompt, pipeline, action, system (optional only for: component, rules).`,
+            `Required for type: skill, tool, agent, prompt, pipeline, system (optional only for: component, rules).`,
             ``,
             `Minimal example:`,
             ``,
@@ -59,7 +59,7 @@ export async function readManifest(
             `    allowed: false            # set true if the package shells out`,
             `  secrets: []                 # env vars or secret keys the package reads`,
             ``,
-            `See: https://github.com/the-metafactory/arc/blob/main/README.md#capabilities`,
+            `See: https://github.com/the-metafactory/arc/blob/main/README.md#capability-declarations`,
           ].join("\n"),
         );
       }
--- a/src/lib/manifest.ts
+++ b/src/lib/manifest.ts
@@ -42,7 +42,25 @@ export async function readManifest(
       // capabilities optional for component and rules types, required for others
       if (!parsed.capabilities && parsed.type !== "component" && parsed.type !== "rules") {
         throw new Error(
-          `Invalid ${filename}: missing required field 'capabilities' (only optional for type: component)`
+          [
+            `Invalid ${filename}: missing required field 'capabilities'`,
+            `Required for type: skill, tool, agent, prompt, pipeline, action, system (optional only for: component, rules).`,
+            ``,
+            `Minimal example:`,
+            ``,
+            `capabilities:`,
+            `  filesystem:`,
+            `    read: []                  # paths the package reads from`,
+            `    write: []                 # paths the package writes to`,
+            `  network:`,
+            `    - domain: api.example.com`,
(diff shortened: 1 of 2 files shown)


jcfischer:
Addressed both factual bugs in commit f9c9cb2.

- ✅ Dropped `action` from the required-types list. Now: `skill, tool, agent, prompt, pipeline, system`.
- ✅ Fixed README anchor to `#capability-declarations` (matches actual heading).
- ➕ Tightened test: asserts message does NOT contain `action` AND link substring is the corrected anchor (not the bare `#capabilities`).

Suite **622/622**, tsc clean.

**Side observation worth a follow-up:** `ArcManifest.type` (src/types.ts:190) lists `skill | system | tool | agent | prompt | component | pipeline | rules | library` but `ArtifactType` (src/types.ts:6) lists `skill | agent | prompt | tool | component | pipeline | rules | library | action`. `action` is supported in practice (artifact-installer.ts handles it) but missing from `ArcManifest.type`. Mismatch worth tightening — will file as separate issue.

PR #96, opened 2026-04-25
fix(install): pre-validate provides.files + rollback on hook gate fail

### LINKED ISSUE #89: install: rollback partial symlinks when provides.files validation fails

jcfischer:
## Summary

Closes #89.

\`install\` previously left orphan symlinks under \`~/.claude/{skills,agents,bin,...}\` when validation failed mid-flow. Two distinct partial-state shapes:

1. \`provides.files\` entry N's source missing → entries 1..N-1 already linked, per-type primary symlinks already in place.
2. Hook gate trips after symlinks landed → symlinks orphaned, \`settings.json\` untouched, DB unaware.

User had no clean recovery path: \`arc remove\` couldn't drive cleanup (no DB entry), so manual rm was the only option.

## Changes

- **\`src/lib/artifact-installer.ts\`** — implement both fixes from the issue:
  1. **Pre-pass validation (option 1).** Before the per-type switch, walk every \`provides.files\` entry and assert each \`source\` exists in the package. If any missing, return immediately with \`filesMissingSource\` populated and **zero filesystem mutations**.
  2. **Rollback tracking (option 2).** Every symlink (per-type primary, CLI bins, \`provides.files\` target) and shim now goes through \`linkTracked\` / \`shimTracked\` helpers that record the path on an \`ArtifactSymlinkRecord\`. New \`rollbackArtifactSymlinks()\` helper unlinks them best-effort.

- **\`src/commands/install.ts\`** — call \`rollbackArtifactSymlinks(symlinkResult.record)\` on the hook-gate failure path. Missing hook target now tears down everything that was placed before the gate ran.

## Tests (test/commands/install.test.ts, +2)

- **Pre-validation:** manifest with two \`provides.files\`, second source missing. Install fails with \`provides.files\` diagnostic. The FIRST entry's target is also absent (pre-validation prevented it being created), and the per-type skill symlink is absent.
- **Hook-gate rollback:** \`provides.files\` succeeds but a separate hook command references an unlanded file. Install fails with hook diagnostic. Skill symlink AND \`provides.files\` target both rolled back.

Suite: 624/624 pass, tsc clean.

### DIFF
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -6,7 +6,12 @@ import { readManifest, readLibraryArtifacts, assessRisk, formatCapabilities } fr
 import { recordInstall, getSkill } from "../lib/db.js";
 import { runScript } from "../lib/scripts.js";
 import { registerHooks, resolveHooksFromManifest, findMissingHookFiles } from "../lib/hooks.js";
-import { createArtifactSymlinks, resolveArtifactSourceDir, installNodeDependencies } from "../lib/artifact-installer.js";
+import {
+  createArtifactSymlinks,
+  resolveArtifactSourceDir,
+  installNodeDependencies,
+  rollbackArtifactSymlinks,
+} from "../lib/artifact-installer.js";
 import { wireExtensions } from "../lib/extensions.js";
 import { extractRepoName } from "../lib/repo-name.js";
 
@@ -288,6 +293,10 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
       const detail = missingHookFiles
         .map((m) => `  - ${m.event}: ${m.command}\n      missing: ${m.missingPath}`)
         .join("\n");
+      // #89: roll back any symlinks/shims createArtifactSymlinks placed before
+      // we hit this gate, so the install fails clean rather than leaving
+      // orphan entries under ~/.claude/{skills,bin,...} for the user to clean.
+      await rollbackArtifactSymlinks(symlinkResult.record);
       return {
         success: false,
         error:
@@ -556,6 +565,10 @@ export async function installSingleArtifact(
       const detail = missingHookFiles
         .map((m) => `  - ${m.event}: ${m.command}\n      missing: ${m.missingPath}`)
         .join("\n");
+      // #89: roll back any symlinks/shims createArtifactSymlinks placed before
+      // we hit this gate, so the install fails clean rather than leaving
+      // orphan entries under ~/.claude/{skills,bin,...} for the user to clean.
+      await rollbackArtifactSymlinks(symlinkResult.record);
       return {
         success: false,
         error:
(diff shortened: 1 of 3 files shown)


jcfischer:
Addressed Luna's review in commit 514bd47.

- ✅ **JSDoc tightened** on `rollbackArtifactSymlinks` — no longer claims postinstall coverage. Now reads: "Currently called by install when the hook-validation gate fails. Postinstall-script failure also leaves partial state but is a strictly larger fix; tracked separately."
- ✅ **Non-ENOENT errors surfaced** via `console.warn` instead of silent `.catch`. ENOENT still tolerated (best-effort across all entries); permission-denied etc. now leave the user aware of orphans.
- ✅ **Hook-gate test tightened** — added `provides.cli` to the fixture so bin symlink + PATH shim are actually created. Test now asserts both are torn down (was vacuous before).
- ➕ **Filed #97** — `install: rollback symlinks + unregister hooks on postinstall script failure`. Labeled `bug, next`. Body cites Luna's review, lists the three rollback components needed (symlinks, hooks unregister, possible DB cleanup), and proposes the call-site changes.

Suite stable at 624/624. Re-pinging.

PR #111, opened 2026-05-07
feat(nats): bot identity provisioning CLI

jcfischer:
## Summary

- Adds \`arc nats\` subcommand group for per-bot NATS user provisioning
- 4 commands: \`add-bot\`, \`reissue-bot\`, \`list-bots\`, \`remove-bot\`
- Wraps NSC with metafactory naming conventions
- Credentials at \`~/.config/nats/{name}.creds\` (mode 600)
- Tested end-to-end: add → list → remove with real NSC

## Architecture decision

Bot identity provisioning is operator-level infrastructure (Layer 2-3 in the [aPaaS vision](https://github.com/the-metafactory/meta-factory/blob/main/research/2026-05-06-apaas-nervous-system-vision.md)). Lives in arc (operator CLI), not grove (observability/dashboard).

## Changes

- \`src/commands/nats.ts\` — new: 4 exported functions + NSC helpers
- \`src/cli.ts\` — wires \`arc nats\` subcommand group

## Test plan

- [x] \`arc nats add-bot spike-test -a OP_JC\` — creates user + creds
- [x] \`arc nats list-bots -a OP_JC\` — shows spike-test in list
- [x] \`arc nats remove-bot spike-test -a OP_JC --delete-creds\` — removes user + file
- [x] \`arc nats list-bots\` (no account) — auto-detects active account

Part of [grove#320](https://github.com/the-metafactory/grove/issues/320) — Task T-3.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -37,6 +37,7 @@ import {
 import type { CatalogEntry, ArtifactType, PackageTier, RegistrySource, SourceType } from "./types.js";
 import { login } from "./commands/login.js";
 import { logout } from "./commands/logout.js";
+import { addBot, reissueBot, listBots, removeBot } from "./commands/nats.js";
 import { bundle, formatBundle } from "./commands/bundle.js";
 import { publish, formatPublish } from "./commands/publish.js";
 import {
@@ -1173,8 +1174,6 @@ catalog
 
 // ── NATS bot identity commands ─────────────────────────────
 
-import { addBot, reissueBot, listBots, removeBot } from "./commands/nats.js";
-
 const nats = program
   .command("nats")
   .description("NATS bot identity management — provision per-bot users");
@@ -1212,8 +1211,9 @@ nats
   .command("remove-bot <name>")
   .description("Revoke a bot user and optionally delete credentials")
   .option("-a, --account <account>", "NSC account name")
+  .option("-o, --output <path>", "Credentials file path to delete (if --delete-creds)")
   .option("--delete-creds", "Also delete the credentials file")
-  .action((name: string, opts: { account?: string; deleteCreds?: boolean }) => {
+  .action((name: string, opts: { account?: string; output?: string; deleteCreds?: boolean }) => {
     removeBot(name, opts);
   });
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1171,4 +1171,50 @@ catalog
     }
   });
 
+// ── NATS bot identity commands ─────────────────────────────
+
+import { addBot, reissueBot, listBots, removeBot } from "./commands/nats.js";
+
+const nats = program
+  .command("nats")
+  .description("NATS bot identity management — provision per-bot users");
+
+nats
+  .command("add-bot <name>")
+  .description("Issue a new per-bot NATS user with credentials")
+  .option("-a, --account <account>", "NSC account name (default: active account)")
+  .option("--pub <subjects>", "Comma-separated publish permissions")
+  .option("--sub <subjects>", "Comma-separated subscribe permissions")
+  .option("-o, --output <path>", "Credentials output path")
+  .option("--force", "Overwrite existing user")
+  .action((name: string, opts: { account?: string; pub?: string; sub?: string; output?: string; force?: boolean }) => {
+    addBot(name, opts);
+  });
+
+nats
+  .command("reissue-bot <name>")
+  .description("Revoke and re-issue credentials for a bot user")
+  .option("-a, --account <account>", "NSC account name")
+  .option("-o, --output <path>", "Credentials output path")
+  .action((name: string, opts: { account?: string; output?: string }) => {
+    reissueBot(name, opts);
+  });
+
+nats
+  .command("list-bots")
+  .description("List bot users under current operator account")
+  .option("-a, --account <account>", "NSC account name")
+  .action((opts: { account?: string }) => {
+    listBots(opts.account);
+  });
+
+nats
+  .command("remove-bot <name>")
+  .description("Revoke a bot user and optionally delete credentials")
+  .option("-a, --account <account>", "NSC account name")
+  .option("--delete-creds", "Also delete the credentials file")
+  .action((name: string, opts: { account?: string; deleteCreds?: boolean }) => {
+    removeBot(name, opts);
+  });
+
 program.parse();
--- a/src/commands/nats.ts
+++ b/src/commands/nats.ts
@@ -3,47 +3,45 @@
  *
  * Wraps NSC to provision per-bot NATS users under an operator's account.
  * Part of grove#320 (bot-level AAA) — operator-level infrastructure tooling.
- *
- * Commands:
- *   arc nats add-bot <name>       Issue a new per-bot NATS user
- *   arc nats reissue-bot <name>   Revoke + re-issue credentials
- *   arc nats list-bots            List bot users under current account
- *   arc nats remove-bot <name>    Revoke a bot user
  */
 
-import { execSync } from "node:child_process";
-import { existsSync, writeFileSync, unlinkSync, mkdirSync } from "node:fs";
+import { existsSync, writeFileSync, unlinkSync, mkdirSync, chmodSync } from "node:fs";
 import { join, dirname } from "node:path";
 import { homedir } from "node:os";
+import { spawnSync } from "node:child_process";
 
 const CREDS_DIR = join(homedir(), ".config", "nats");
 const NAMING_RE = /^[a-z][a-z0-9-]*$/;
+const NATS_SUBJECT_RE = /^[a-zA-Z0-9.*>_-]+$/;
 
-function nsc(args: string): string {
-  try {
-    // nsc writes some output to stderr (e.g., `nsc env`), merge both streams
-    return execSync(`nsc ${args} 2>&1`, { encoding: "utf-8" }).trim();
-  } catch (e: unknown) {
-    const err = e as { stderr?: string; stdout?: string };
-    const msg = err.stderr?.trim() || err.stdout?.trim() || "unknown error";
-    throw new Error(`nsc ${args.split(" ")[0]} failed: ${msg}`);
+function nsc(args: string[]): string {
+  const result = spawnSync("nsc", args, { encoding: "utf-8" });
+  if (result.status !== 0) {
+    const msg = (result.stderr || result.stdout || "unknown error").trim();
+    throw new Error(`nsc ${args[0]} failed: ${msg}`);
   }
+  // nsc writes some output to stderr (e.g., `nsc env`)
+  return (result.stdout + result.stderr).trim();
 }
 
 export function ensureNscInstalled(): void {
-  try {
-    execSync("which nsc", { encoding: "utf-8" });
-  } catch {
+  const result = spawnSync("which", ["nsc"], { encoding: "utf-8" });
+  if (result.status !== 0) {
     console.error("Error: nsc not found on PATH.");
     console.error("Install: brew install nats-io/nats-tools/nsc");
-    console.error("  or: https://github.com/nats-io/nsc");
     process.exit(1);
   }
 }
 
+function validateSubject(subject: string): void {
+  if (!NATS_SUBJECT_RE.test(subject)) {
+    throw new Error(`Invalid NATS subject: "${subject}" — only alphanumeric, dots, wildcards, hyphens, underscores allowed`);
+  }
+}
+
 export function detectAccount(): string {
-  const env = nsc("env");
-  const match = env.match(/Current Account\s+\|[^|]*\|\s+(\S+)/);
+  const output = nsc(["env"]);
+  const match = output.match(/Current Account\s+\|[^|]*\|\s+(\S+)/);
   if (match) return match[1];
   throw new Error("Cannot detect NSC account. Run: nsc env -a <account>");
 }
@@ -54,11 +52,24 @@ function credsPath(name: string): string {
 
 function userExists(account: string, name: string): boolean {
   try {
-    nsc(`describe user -a ${account} -n ${name}`);
+    nsc(["describe", "user", "-a", account, "-n", name]);
     return true;
   } catch {
     return false;
   }
+}
+
+function ensureCredsDir(path: string): void {
+  const dir = dirname(path);
+  if (!existsSync(dir)) {
+    mkdirSync(dir, { recursive: true, mode: 0o700 });
+  }
+  chmodSync(dir, 0o700);
+}
+
+function writeCredsFile(path: string, content: string): void {
+  writeFileSync(path, content, { mode: 0o600 });
+  chmodSync(path, 0o600);
 }
 
 export interface AddBotOptions {
@@ -90,31 +101,32 @@ export function addBot(name: string, opts: AddBotOptions): void {
       console.error(`Error: user "${name}" exists under "${account}". Use --force to re-create.`);
       process.exit(1);
     }
-    nsc(`delete user -a ${account} -n ${name}`);
+    nsc(["delete", "user", "-a", account, "-n", name]);
     console.log(`Removed existing user: ${name}`);
   }
 
-  nsc(`add user -a ${account} -n ${name}`);
+  nsc(["add", "user", "-a", account, "-n", name]);
   console.log(`Created NATS user: ${name} (account: ${account})`);
 
   if (opts.pub) {
     for (const subj of opts.pub.split(",").map((s) => s.trim())) {
-      nsc(`edit user -a ${account} -n ${name} --allow-pub "${subj}"`);
+      validateSubject(subj);
+      nsc(["edit", "user", "-a", account, "-n", name, "--allow-pub", subj]);
     }
     console.log(`  publish: ${opts.pub}`);
   }
 
   if (opts.sub) {
     for (const subj of opts.sub.split(",").map((s) => s.trim())) {
-      nsc(`edit user -a ${account} -n ${name} --allow-sub "${subj}"`);
+      validateSubject(subj);
+      nsc(["edit", "user", "-a", account, "-n", name, "--allow-sub", subj]);
     }
     console.log(`  subscribe: ${opts.sub}`);
   }
 
-  const creds = nsc(`generate creds -a ${account} -n ${name}`);
-  const outDir = dirname(outPath);
-  if (!existsSync(outDir)) mkdirSync(outDir, { recursive: true });
-  writeFileSync(outPath, creds, { mode: 0o600 });
+  const creds = nsc(["generate", "creds", "-a", account, "-n", name]);
+  ensureCredsDir(outPath);
+  writeCredsFile(outPath, creds);
   console.log(`  credentials: ${outPath} (mode 600)`);
 }
 
@@ -133,11 +145,28 @@ export function reissueBot(name: string, opts: ReissueBotOptions): void {
     process.exit(1);
   }
 
-  nsc(`delete user -a ${account} -n ${name}`);
-  nsc(`add user -a ${account} -n ${name}`);
+  // Capture existing user config before delete for rollback
+  let existingConfig: string;
+  try {
+    existingConfig = nsc(["describe", "user", "-a", account, "-n", name, "--json"]);
+  } catch {
+    console.error(`Error: cannot read existing user config for "${name}".`);
+    process.exit(1);
+  }
 
-  const creds = nsc(`generate creds -a ${account} -n ${name}`);
-  writeFileSync(outPath, creds, { mode: 0o600 });
+  nsc(["delete", "user", "-a", account, "-n", name]);
+
+  try {
+    nsc(["add", "user", "-a", account, "-n", name]);
+  } catch (err) {
+    console.error(`Error: failed to re-create user "${name}". Previous config saved to stderr.`);
+    process.stderr.write(`Previous user config:\n${existingConfig}\n`);
+    process.exit(1);
+  }
+
+  const creds = nsc(["generate", "creds", "-a", account, "-n", name]);
+  ensureCredsDir(outPath);
+  writeCredsFile(outPath, creds);
 
   console.log(`Re-issued credentials for ${name}`);
   console.log(`  credentials: ${outPath} (mode 600)`);
@@ -147,12 +176,13 @@ export function reissueBot(name: string, opts: ReissueBotOptions): void {
 export function listBots(account?: string): void {
   ensureNscInstalled();
   const acct = account ?? detectAccount();
-  console.log(nsc(`list users -a ${acct}`));
+  console.log(nsc(["list", "users", "-a", acct]));
 }
 
 export interface RemoveBotOptions {
   account?: string;
   deleteCreds?: boolean;
+  output?: string;
 }
 
 export function removeBot(name: string, opts: RemoveBotOptions): void {
@@ -164,14 +194,16 @@ export function removeBot(name: string, opts: RemoveBotOptions): void {
     process.exit(1);
   }
 
-  nsc(`delete user -a ${account} -n ${name}`);
+  nsc(["delete", "user", "-a", account, "-n", name]);
   console.log(`Removed user: ${name} (account: ${account})`);
 
   if (opts.deleteCreds) {
-    const path = credsPath(name);
+    const path = opts.output ?? credsPath(name);
     if (existsSync(path)) {
       unlinkSync(path);
       console.log(`Deleted credentials: ${path}`);
+    } else {
+      console.log(`Warning: no credentials file at ${path} (custom -o path used at creation?)`);
     }
   }
 }
--- a/src/commands/nats.ts
+++ b/src/commands/nats.ts
@@ -22,13 +22,22 @@ const NSC_CONFIG_CANDIDATES = [
 
 function nsc(args: string[]): string {
   const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
-  const stdout = result.stdout.toString();
-  const stderr = result.stderr.toString();
+  const stdout = result.stdout.toString().trim();
+  const stderr = result.stderr.toString().trim();
   if (result.exitCode !== 0) {
-    const msg = (stderr || stdout || "unknown error").trim();
-    throw new Error(`nsc ${args[0]} failed: ${msg}`);
+    throw new Error(`nsc ${args[0]} failed: ${stderr || stdout || "unknown error"}`);
   }
-  return (stdout + stderr).trim();
+  return stdout;
+}
+
+function nscWithStderr(args: string[]): string {
+  const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
+  if (result.exitCode !== 0) {
+    const stderr = result.stderr.toString().trim();
+    const stdout = result.stdout.toString().trim();
+    throw new Error(`nsc ${args[0]} failed: ${stderr || stdout || "unknown error"}`);
+  }
+  return (result.stdout.toString() + result.stderr.toString()).trim();
 }
 
 export function ensureNscInstalled(): void {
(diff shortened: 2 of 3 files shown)


src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+import { join, dirname } from "node:path";
+import { homedir } from "node:os";
+
+const CREDS_DIR = join(homedir(), ".config", "nats");
+const NAMING_RE = /^[a-z][a-z0-9-]*$/;
+
+function nsc(args: string): string {
+  try {
+    // nsc writes some output to stderr (e.g., `nsc env`), merge both streams
+    return execSync(`nsc ${args} 2>&1`, { encoding: "utf-8" }).trim();
```
mellanon: **[critical/security]** Shell command injection. `nsc()` runs `execSync(`nsc ${args} 2>&1`)` with template-string interpolation. The bot `name` is gated by `NAMING_RE`, but `account` (from `--account`) and `subj` (from `--pub`/`--sub` items) flow straight in unvalidated. `arc nats add-bot foo -a 'OP_JC; rm -rf ~'` runs the second command. The repo idiom — `Bun.spawnSync(['nsc', 'add', 'user', ...args])` (see `src/commands/catalog.ts`, `src/commands/self-update.ts`) — uses an argv array, no shell, no injection.

Also: the inline `2>&1` shell redirect inside the template means any deprecation/warning nsc writes to stderr during `generate creds` is concatenated to the .creds payload and written to disk. Silent corruption.

**Fix:** rewrite `nsc()` on top of `Bun.spawnSync` with separate stdout/stderr streams; drop the shell entirely. Validate `account` against the same regex as `name`.

src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+    nsc(`delete user -a ${account} -n ${name}`);
+    console.log(`Removed existing user: ${name}`);
+  }
+
+  nsc(`add user -a ${account} -n ${name}`);
+  console.log(`Created NATS user: ${name} (account: ${account})`);
+
+  if (opts.pub) {
+    for (const subj of opts.pub.split(",").map((s) => s.trim())) {
+      nsc(`edit user -a ${account} -n ${name} --allow-pub "${subj}"`);
```
mellanon: **[critical/security]** `subj` is unvalidated and shell-interpolated. `arc nats add-bot foo --pub 'a.b,";rm -rf ~/;"'` breaks out of the double quotes inside the template. Same fix as the `nsc()` helper — use `Bun.spawnSync` with an argv array. Then `subj` becomes a single argv element and shell metacharacters are inert.

src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+    for (const subj of opts.sub.split(",").map((s) => s.trim())) {
+      nsc(`edit user -a ${account} -n ${name} --allow-sub "${subj}"`);
+    }
+    console.log(`  subscribe: ${opts.sub}`);
+  }
+
+  const creds = nsc(`generate creds -a ${account} -n ${name}`);
+  const outDir = dirname(outPath);
+  if (!existsSync(outDir)) mkdirSync(outDir, { recursive: true });
+  writeFileSync(outPath, creds, { mode: 0o600 });
```
mellanon: **[major/security]** `writeFileSync(path, data, { mode: 0o600 })` only applies the mode when the file is **created**. If the file already exists (e.g. mid-overwrite path, or user passes `-o /existing/file`), the existing mode is preserved. Combined with the `existsSync(outPath) && !opts.force` check above (line 83), `--force` lets you overwrite a 0644 file and the creds land at 0644.

**Fix:** open with the `wx` flag (exclusive create) after unlinking on `--force`, OR explicitly `chmodSync(outPath, 0o600)` after write.

src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+  if (opts.sub) {
+    for (const subj of opts.sub.split(",").map((s) => s.trim())) {
+      nsc(`edit user -a ${account} -n ${name} --allow-sub "${subj}"`);
+    }
+    console.log(`  subscribe: ${opts.sub}`);
+  }
+
+  const creds = nsc(`generate creds -a ${account} -n ${name}`);
+  const outDir = dirname(outPath);
+  if (!existsSync(outDir)) mkdirSync(outDir, { recursive: true });
```
mellanon: **[major/security]** `mkdirSync(outDir, { recursive: true })` doesn't restrict directory mode — `~/.config/nats/` ends up at whatever umask gives, typically 0755. Other local users can then `ls` to enumerate every bot identity you've ever provisioned. Pass `mode: 0o700` and consider `chmodSync(outDir, 0o700)` for the case where the dir already exists with a permissive mode.

src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+  const account = opts.account ?? detectAccount();
+  const outPath = opts.output ?? credsPath(name);
+
+  if (!userExists(account, name)) {
+    console.error(`Error: user "${name}" not found under "${account}".`);
+    process.exit(1);
+  }
+
+  nsc(`delete user -a ${account} -n ${name}`);
+  nsc(`add user -a ${account} -n ${name}`);
```
mellanon: **[major/quality]** `reissueBot` deletes the user before re-creating it. If `add user` (line 137) fails for any reason — nsc bug, permission change, transient — you've destroyed the user with no way back. `addBot` has the same gap inside its own `--force` branch (line 93).

**Fix:** capture the existing JWT (`nsc describe user --json`) before delete, and on failure restore it. Or: gate `reissueBot` behind a confirmation prompt / `--force`, since this is a one-way door for the live bot.

Also: `reissueBot` doesn't run `NAMING_RE.test(name)` (unlike `addBot`) and overwrites `outPath` without the `existsSync && !force` check that `addBot` does. Inconsistent.

src/commands/nats.ts
mellanon: **[major/quality]** Partial-state failure. If the `--allow-pub` loop succeeds for items 1..n-1 and then nsc fails on item n, you have a user with partial publish permissions and no creds file. No rollback. Same for `--sub` (lines 107-112). And if the loops succeed but `generate creds` (line 114) fails, the user exists with permissions but no creds — and re-running `add-bot` without `--force` errors out.

**Fix:** at minimum, wrap from line 97 (`add user`) onward in try/catch and `nsc delete user` on any failure before re-throwing. Better: build the user with all permissions in one nsc invocation if nsc supports it.

src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  if (!userExists(account, name)) {
+    console.error(`Error: user "${name}" not found under "${account}".`);
+    process.exit(1);
+  }
+
+  nsc(`delete user -a ${account} -n ${name}`);
+  console.log(`Removed user: ${name} (account: ${account})`);
+
+  if (opts.deleteCreds) {
```
mellanon, line 248: **[major/quality]** `removeBot --delete-creds` always deletes `credsPath(name)` (the default `~/.config/nats/{name}.creds`). If the user originally provisioned with `addBot ... -o /custom/path/foo.creds`, the custom file is left orphaned on disk while the NATS user is gone. Either accept `-o` here too, or warn that `--delete-creds` only removes the default location.

src/commands/nats.ts
```diff
@@ -0,0 +1,177 @@
...
+    console.error("Error: nsc not found on PATH.");
+    console.error("Install: brew install nats-io/nats-tools/nsc");
+    console.error("  or: https://github.com/nats-io/nsc");
+    process.exit(1);
+  }
+}
+
+export function detectAccount(): string {
+  const env = nsc("env");
+  const match = env.match(/Current Account\s+\|[^|]*\|\s+(\S+)/);
```
mellanon: **[warning/quality]** Brittle parser. Regexing the `nsc env` table (`Current Account\s+\|[^|]*\|\s+(\S+)`) breaks the moment NSC tweaks its column layout. Many `nsc` subcommands accept structured output; check whether a JSON form exists for `env` and use that, or invoke the source-of-truth file directly (`~/.nsc/...`). Failing that, capture the current expected-format string in a test fixture so format drift is caught.

src/cli.ts
```diff
@@ -1171,4 +1171,50 @@ catalog
     }
   });
 
+// ── NATS bot identity commands ─────────────────────────────
+
```
mellanon, line 1176: **[warning/architecture]** Mid-file `import` statement. Every other command in arc declares its imports at the top of `src/cli.ts`. Slipping this one in at line 1175 to keep the diff localized hurts readability and breaks the convention. Move it up with the other command imports.

mellanon (changes requested):
Full review (6 lenses + duplication) — shell injection in nsc() wrapper is the headline; stderr-merged creds, partial-state failure paths, weak permissions hygiene, and zero automated tests round it out.

verdict: blockers=2 majors=5 nits=2 — recommend: request-changes

src/commands/nats.ts
```diff
@@ -0,0 +1,209 @@
...
+      validateSubject(subj);
+      nsc(["edit", "user", "-a", account, "-n", name, "--allow-sub", subj]);
+    }
+    console.log(`  subscribe: ${opts.sub}`);
+  }
+
+  const creds = nsc(["generate", "creds", "-a", account, "-n", name]);
+  ensureCredsDir(outPath);
+  writeCredsFile(outPath, creds);
+  console.log(`  credentials: ${outPath} (mode 600)`);
```
mellanon, line 170: **[major/quality]** Partial-state failure still unaddressed. Once `nsc add user` succeeds at line 108, there is no try/catch around the permission loops or `generate creds`. If `--allow-sub` fails on item 3 of 5, you end up with a user holding partial publish and partial subscribe and no creds file. If `generate creds` (line 127) fails, the user exists with permissions but no creds — and re-running without `--force` errors out at line 99.

Fix: wrap from line 108 onward in try/catch; on any failure, `nsc delete user` so nsc state is left clean and the operator can retry. This is the same shape as the reissue rollback story — see next comment.

src/commands/nats.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  try {
+    nsc(["add", "user", "-a", account, "-n", name]);
+  } catch (err) {
+    console.error(`Error: failed to re-create user "${name}". Previous config saved to stderr.`);
+    process.stderr.write(`Previous user config:\n${existingConfig}\n`);
+    process.exit(1);
+  }
+
+  const creds = nsc(["generate", "creds", "-a", account, "-n", name]);
```
mellanon: **[major/quality]** Rollback is informational, not actual restoration. `describe user --json` (line 151) captures the user metadata — name, public key, claim payload — but **not** the JWT itself. There is no recipe by which an operator can reconstruct the live user from that JSON blob. On `add user` failure (line 160), the user is destroyed and stderr gets a dump that is not directly actionable.

Two real fixes:
- Back up the JWT file at `~/.nsc/nats/{operator}/accounts/{account}/users/{name}.jwt` before `nsc delete user` on line 157, and on `add user` failure copy it back; or
- Wire the failure path into a documented recovery: which `nsc add user --tag` flags reconstruct from the JSON, and emit the exact command in the error message.

As written this is a tombstone with the corpse's name on it, not a rollback.

src/commands/nats.ts
```diff
@@ -0,0 +1,209 @@
...
+
+function validateSubject(subject: string): void {
+  if (!NATS_SUBJECT_RE.test(subject)) {
+    throw new Error(`Invalid NATS subject: "${subject}" — only alphanumeric, dots, wildcards, hyphens, underscores allowed`);
+  }
+}
+
+export function detectAccount(): string {
+  const output = nsc(["env"]);
+  const match = output.match(/Current Account\s+\|[^|]*\|\s+(\S+)/);
```
mellanon, line 79: **[warning/quality]** Brittle parser — unchanged from prior round. `nsc env`'s table layout is the regex's load-bearing assumption; one column tweak upstream and `detectAccount` silently fails over to the throw at line 46. Options:
- Probe for a structured form first (`nsc env --json`) and fall back to the regex.
- Read NSC's own config at `~/.nsc/.nsc.json` directly.
- At minimum, capture the current expected `nsc env` output as a test fixture so format drift is caught at CI time rather than at 2am after an nsc upgrade.

Not a blocker. Logging it again so it doesn't age out.

src/commands/nats.ts
```diff
@@ -0,0 +1,209 @@
+/**
```
mellanon, line 1: **[major/quality]** No tests for `src/commands/nats.ts`. Every other command in the repo has a matching `test/commands/*.test.ts` — twenty test files for twenty commands (audit, bundle, catalog, install, login, logout, publish, …). This PR adds a 209-line command module with zero. Manual end-to-end in the PR body proves the happy path works once; tests prove it keeps working as nsc and the rest of arc evolve.

Minimum: unit tests for `validateSubject` (boundary cases — `*`, `>`, embedded shell metachars), `detectAccount` against a captured `nsc env` fixture, and the partial-state failure paths in `addBot`/`reissueBot`.

mellanon (changes requested):
## Re-review of the-metafactory/arc#111

Both criticals from the prior round are properly resolved — `spawnSync` argv pattern eliminates the shell injection class, and `validateSubject` adds defense in depth. File and directory permissions are tightened correctly. The mid-file import is back where it belongs.

What's still open:

| Prior finding | Status |
|---|---|
| critical/security: shell injection in `nsc()` helper | ✅ resolved (spawnSync argv) |
| critical/security: `subj` shell-interpolated | ✅ resolved (argv + validateSubject) |
| major/security: writeFileSync mode-on-create only | ✅ resolved (chmodSync after write) |
| major/security: mkdirSync 0o755 default | ✅ resolved (mode 0o700 + chmod) |
| major/quality: reissueBot rollback | ❌ **partial** — informational dump, not restoration |
| major/quality: addBot partial-state failure | ❌ **unaddressed** — no try/catch around perm loops + creds gen |
| major/quality: removeBot --delete-creds path | ✅ resolved (-o flag + warning) |
| warning/quality: brittle nsc env parser | ❌ **unaddressed** — same regex |
| warning/architecture: mid-file import | ✅ resolved |

Plus a major from the prior summary that wasn't inline-flagged but I'm calling out now: still no `test/commands/nats.test.ts`. Twenty other commands have matching tests.

verdict: blockers=0 majors=3 nits=1 — recommend: request-changes

src/commands/nats.ts
```diff
@@ -0,0 +1,222 @@
...
+    return false;
+  }
+}
+
+function ensureCredsDir(path: string): void {
+  const dir = dirname(path);
+  if (!existsSync(dir)) {
+    mkdirSync(dir, { recursive: true, mode: 0o700 });
+  }
+  chmodSync(dir, 0o700);
```
mellanon: **[major/security]** `ensureCredsDir` unconditionally `chmodSync(dir, 0o700)` on `dirname(outPath)` — even when `outPath` is user-supplied via `-o`. `arc nats add-bot foo -o ~/Documents/foo.creds` will silently chmod `~/Documents` to 0700. Same for `arc nats reissue-bot ... -o /shared/dir/foo.creds`.

The cycle-2 fix was correct for the *default* `~/.config/nats/` path but is now too broad: it asserts perms on a directory the user owns and explicitly chose. The chmod is also persistent — there's no logging that we changed perms.

**Fix:** only chmod the dir when it's the default `CREDS_DIR`. For custom `-o` paths, either skip the chmod or print a warning. Or: bound the perm tightening to the file itself (which `writeCredsFile` already does via `chmodSync(path, 0o600)`). Most robust: write to a per-user-owned dir only, and reject `-o` paths whose parent has non-user ownership.

src/commands/nats.ts
```diff
@@ -0,0 +1,222 @@
...
+ * Part of grove#320 (bot-level AAA) — operator-level infrastructure tooling.
+ */
+
+import { existsSync, readFileSync, writeFileSync, unlinkSync, mkdirSync, chmodSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { homedir } from "node:os";
+import { spawnSync } from "node:child_process";
+
+const CREDS_DIR = join(homedir(), ".config", "nats");
+const NSC_CONFIG = join(homedir(), ".config", "nats", "nsc", "nsc.json");
```
mellanon: **[major/quality]** The "primary" config path `~/.config/nats/nsc/nsc.json` does not match where NSC actually stores its config. Default NSC layout is `~/.nsc/nsc.json`, or `${XDG_CONFIG_HOME}/nsc/nsc.json` (typically `~/.config/nsc/nsc.json`) on newer builds. `~/.config/nats/nsc/nsc.json` is neither — it sits inside the *creds* directory this CLI manages, which NSC has no reason to write to.

On a default install, `existsSync(NSC_CONFIG)` returns false and `detectAccount` falls through to the regex parser every time. Verified on this box: `~/.config/nats/` exists with creds files but no `nsc/` subdir; `~/.nsc` and `~/.config/nsc` don't exist.

Net: the brittle-parser finding from the prior round isn't actually addressed — the regex is still the load-bearing path. Either:
- read `~/.nsc/nsc.json` (with XDG fallback to `~/.config/nsc/nsc.json`), or
- use `nsc env --json` if a structured form exists, or
- accept the regex as the only path and pin a fixture test so format drift breaks CI rather than ops at 2am.

If you've got a real machine where NSC writes `~/.config/nats/nsc/nsc.json`, share the `nsc` version + how it was initialised and we'll trace the divergence.

src/commands/nats.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  try {
+    nsc(["add", "user", "-a", account, "-n", name]);
+  } catch (err) {
+    console.error(`CRITICAL: failed to re-create user "${name}" after delete.`);
+    console.error(`The user has been removed but could not be re-created.`);
+    console.error(`Manual recovery: nsc add user -a ${account} -n ${name}`);
+    console.error(`Cause: ${err instanceof Error ? err.message : String(err)}`);
+    process.exit(1);
+  }
```
mellanon, line 213: **[major/quality]** Cycle 3 added rollback for `addBot`, but `reissueBot` is unchanged from cycle 2: `delete user → add user → on failure, log CRITICAL and exit`. There is no rollback. The user is gone, the old creds JWT is invalid forever, and the operator has `nsc add user -a ... -n ...` printed to stderr — which only re-creates a user with a *different* keypair, not the original. The live bot is dead until the operator re-issues creds and pushes them to the bot.

The window where this matters is small (nsc `add user` rarely fails after `delete` succeeds) but the blast radius is total when it does — a bot in production loses its identity with no recovery.

**Fix options:**
- Capture the user's JWT from `~/.nsc/nats/{operator}/accounts/{account}/users/{name}.jwt` before `nsc delete user`, and on `add user` failure restore it via `nsc import user` or by writing the file back.
- Or: gate `reissueBot` behind `--force` and emit a hard warning that this is a one-way door.
- At minimum, persist the recovery path (the JWT we *had*) to a tmp file and reference it in the CRITICAL message so the operator can recover without keypair regeneration.

Flagging again because cycle 3's commit message claims "rollback safety" but only addresses `addBot`.

src/commands/nats.ts
```diff
@@ -0,0 +1,222 @@
...
+ * arc nats — NATS bot identity management.
+ *
+ * Wraps NSC to provision per-bot NATS users under an operator's account.
+ * Part of grove#320 (bot-level AAA) — operator-level infrastructure tooling.
+ */
+
+import { existsSync, readFileSync, writeFileSync, unlinkSync, mkdirSync, chmodSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { homedir } from "node:os";
+import { spawnSync } from "node:child_process";
```
mellanon: **[warning/architecture]** This file uses `spawnSync` from `node:child_process`. Every other command in arc — `catalog.ts`, `self-update.ts`, `upgrade.ts`, `verify.ts`, `info.ts`, `install.ts`, plus libs (`bundle.ts`, `scripts.ts`, `registry-install.ts`, `cosign.ts`, `remote-registry.ts`, `artifact-installer.ts`) — uses `Bun.spawnSync`. ~50 call sites repo-wide, zero of them `node:child_process`.

CLAUDE.md `# Bun` section says: *"Default to using Bun instead of Node.js."* This is the only file that opts out.

**Fix:** swap to `Bun.spawnSync(['nsc', ...args], { stdout: 'pipe', stderr: 'pipe' })` and use the same return-shape (`exitCode`, `stdout`/`stderr` as Buffer/Uint8Array) as the rest of the codebase. Net effect: pattern consistency, no behavioral change.

test/commands/nats.test.ts
```diff
@@ -0,0 +1,84 @@
...
+const NSC_AVAILABLE = spawnSync("which", ["nsc"], { encoding: "utf-8" }).status === 0;
+const TEST_ACCOUNT = "OP_JC";
+const TEST_BOT = "arc-test-bot";
+const CREDS_PATH = join(homedir(), ".config", "nats", `${TEST_BOT}.creds`);
+
+// Cleanup in case prior run left state — use spawnSync to avoid process.exit
+function cleanupTestBot(): void {
+  spawnSync("nsc", ["delete", "user", "-a", TEST_ACCOUNT, "-n", TEST_BOT], { encoding: "utf-8" });
+  const p = join(homedir(), ".config", "nats", `${TEST_BOT}.creds`);
+  try { require("fs").unlinkSync(p); } catch { /* ok */ }
```
mellanon: **[warning/quality]** `require("fs").unlinkSync(p)` mid-function in an ESM file. The existing imports are `existsSync, readFileSync, statSync` — `unlinkSync` simply isn't imported. Either add it to the import on line 4 or use `Bun.spawnSync(["rm", "-f", p])` to match the rest of the codebase.

Minor, but `require` in an ESM module is a code smell and will trip a future ESLint config.

mellanon (changes requested):
Cycle 3 review (6 lenses + duplication) — partial-state rollback on addBot landed cleanly and the test file is welcome. But the cycle 2 "fix dir perms" patch overshoots into user-supplied -o paths, the "config-based account detection" reads from a path NSC doesn't actually use (regex fallback is still load-bearing), and reissueBot's one-way-door is unchanged. One pattern violation: this file is the only one in arc using node:child_process instead of Bun.spawnSync.

verdict: blockers=0 majors=3 nits=2 — recommend: request-changes

src/commands/nats.ts
```diff
@@ -0,0 +1,243 @@
...
+
+function nsc(args: string[]): string {
+  const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
+  const stdout = result.stdout.toString();
+  const stderr = result.stderr.toString();
+  if (result.exitCode !== 0) {
+    const msg = (stderr || stdout || "unknown error").trim();
+    throw new Error(`nsc ${args[0]} failed: ${msg}`);
+  }
+  return (stdout + stderr).trim();
```
mellanon: **[major/security]** `return (stdout + stderr).trim()` is the same class of bug as cycle 1's `2>&1` critical, just relocated from shell to JS-land. `nsc generate creds` writes the credential block to stdout AND status lines to stderr — concatenating both means any nsc info/warning gets baked into the .creds file written one frame up the call stack (`addBot` line 142, `reissueBot` line 190).

The lifecycle test on line 50 only asserts the JWT/seed markers are *contained* in the file — it would pass even with garbage trailing the seed block, so the test does not protect against this regression.

Fix: `return stdout.trim();`. The error path (line 28) already uses stderr correctly. There is no reason to fold stderr into success output.

src/commands/nats.ts
```diff
@@ -0,0 +1,243 @@
...
+  ensureNscInstalled();
+  const account = opts.account ?? detectAccount();
+  const outPath = opts.output ?? defaultCredsPath(name);
+
+  if (!userExists(account, name)) {
+    console.error(`Error: user "${name}" not found under "${account}".`);
+    process.exit(1);
+  }
+
+  // Generate new creds BEFORE deleting old user — validate the operation succeeds
```
mellanon: **[nit/quality]** This comment claims "Generate new creds BEFORE deleting old user — validate the operation succeeds before making destructive changes" — but the code below does the opposite: file-copy backup, then `nsc delete user` (line 185), then `nsc add user` (line 188). The comment then concedes "nsc requires delete+add for new keys" which contradicts the opening claim.

Either rewrite the comment to match what the code actually does (file-level backup as a paper trail; true rollback is unreachable because NSC revocation is one-way), or delete it. Right now it misleads the reader.

src/commands/nats.ts
```diff
@@ -0,0 +1,243 @@
...
+  if (opts.sub) console.log(`  subscribe: ${opts.sub}`);
+  console.log(`  credentials: ${outPath} (mode 600)`);
+}
+
+export interface ReissueBotOptions {
+  account?: string;
+  output?: string;
+}
+
+export function reissueBot(name: string, opts: ReissueBotOptions): void {
```
mellanon, line 178: **[nit/quality]** `addBot` validates `name` against `NAMING_RE` at line 110, but `reissueBot` (here), `removeBot` (line 232), and `listBots` (line 222) skip that check. With `Bun.spawnSync` argv there is no shell injection risk, but a malformed name still produces a cryptic NSC error rather than the clean `bot name "X" must be lowercase alphanumeric + hyphens.` message users get from add-bot. Either extract a `validateBotName(name)` helper and call it in all four entry points, or accept the asymmetry and leave a comment saying so.

mellanon (changes requested):
## Cycle 4 re-review of the-metafactory/arc#111

Cycle 3's open items mostly land cleanly. The chmod scoping is correct (default dir only — `-o` paths get file-mode 0o600 but no parent-dir mode coercion). `Bun.spawnSync` is now uniform across `nats.ts` and the test file; no `node:child_process` left. The reissueBot backup-then-CRITICAL-message is the right shape given NSC's revocation is one-way — the JWT cannot be un-revoked, so this is a paper trail, not rollback, and that's accepted. Multi-candidate NSC config detection plus regex fallback covers reasonable layouts.

| Cycle 3 finding | Status |
|---|---|
| chmod overshoots into `-o` paths | ✅ resolved (default dir gated on `dirname(path) === DEFAULT_CREDS_DIR`) |
| `node:child_process` pattern violation | ✅ resolved (Bun.spawnSync everywhere) |
| reissueBot one-way-door | ✅ accepted (file-level backup + loud CRITICAL msg; true rollback unreachable) |
| nsc.json path candidates | ✅ resolved (3 candidates + regex fallback) |
| test imports | ✅ resolved |

What's new / still open:

- **major/security:** `nsc()` returns `(stdout + stderr).trim()` on success — same class as cycle 1's `2>&1` critical, just moved into JS. Any nsc status line on stderr ends up concatenated into the .creds file at `addBot:142` / `reissueBot:190`. The lifecycle test only checks `BEGIN ... contains` markers so it doesn't catch trailing garbage. Trivial fix: `return stdout.trim();`.
- **nit/quality:** reissueBot's "Generate new creds BEFORE deleting old user" comment (line 175) misrepresents what the code does — the code backs up the file, then deletes, then re-creates. Either rewrite or delete.
- **nit/quality:** `reissueBot`/`removeBot`/`listBots` skip the `NAMING_RE` check that `addBot` runs. No injection risk under argv-spawn, but error UX diverges. Extract a helper.

Not blocking, but worth tracking for follow-up: F-009 is bot-level AAA — provisioning/revocation events should be auditable. Even an append-only line per command would close the loop. CLAUDE.md commands table is still missing `nats.ts` (carry-over nit from cycle 1).

mellanon (approved):
Full review (6 lenses + duplication) — cycle 5 closes the cycle-4 findings cleanly.

**Verified this cycle:**
- `nsc()` returns stdout only — no more stderr status lines bleeding into credentials. The split into `nsc()` (safe, credential-bearing) and `nscWithStderr()` (env parsing only, called from `detectAccount`) is the right shape.
- `validateBotName` now applied uniformly to `reissueBot` and `removeBot`. `listBots` doesn't take a bot name; correctly skipped.
- Comment in `reissueBot` rewritten to match what the code actually does ("Back up old creds file before the destructive delete"), instead of asserting an order NSC doesn't allow.

**Known NSC limitations, acknowledged not bugs:**
- `reissueBot` delete-then-add and `addBot --force` overwrite both destroy the existing user before re-adding. NSC has no atomic rekey; the `.bak` is informational and the CRITICAL log says so.
- `detectAccount` regex on the `nsc env` table is the fallback path — JSON-config detection runs first.

Five cycles is a lot, but every cycle landed substantive fixes. The diff held up.

PR #113, opened 2026-05-07
feat(nats): --with-identity flag + identity export/import commands

jcfischer:
## Summary
- `arc nats add-bot --with-identity`: one command for NATS creds + Ed25519 signing key + principal registration
- `arc identity generate/list/export/import`: full signing identity lifecycle
- Simplifies AAA Phase 1 migration from 7 manual steps (openssl, hand-edit JSON) to 3 commands

## Usage
```bash
# Provision bot with full identity
arc nats add-bot jc-pilot -a OP_JC --with-identity

# Share your principals with another operator
arc identity export > jc-principals.json

# Import another operator's principals
arc identity import andreas-principals.json

# List all registered principals
arc identity list
```

## Closes
arc#112

## Test plan
- [x] `bun test test/commands/identity.test.ts` — 5 pass
- [x] `bun test test/commands/nats.test.ts` — 5 pass (no regressions)
- [x] Key generation produces valid Ed25519 keypairs
- [x] principals.json format matches myelin PrincipalRegistryFile spec
- [x] Import merges correctly (add new, update changed, skip identical)

### DIFF
--- a/src/commands/identity.ts
+++ b/src/commands/identity.ts
@@ -15,7 +15,8 @@ const CONFIG_BASE = process.env.METAFACTORY_CONFIG_DIR ?? join(homedir(), ".conf
 const KEYS_DIR = join(CONFIG_BASE, "keys");
 const REGISTRY_PATH = join(CONFIG_BASE, "principals.json");
 const DID_RE = /^did:mf:[a-z][a-z0-9._-]+$/;
-const NAMING_RE = /^[a-z][a-z0-9-]*$/;
+const BASE64_RE = /^[A-Za-z0-9+/]+=*$/;
+const NAMING_RE = /^[a-z](?:[a-z0-9]|-(?=[a-z0-9]))*$/;
 
 export interface Principal {
   id: string;
@@ -44,6 +45,17 @@ function keyPath(name: string): string {
   return join(KEYS_DIR, `${name}.key`);
 }
 
+function validatePrincipal(p: unknown, index: number): asserts p is Principal {
+  if (!p || typeof p !== "object") throw new Error(`principals[${index}]: not an object`);
+  const r = p as Record<string, unknown>;
+  if (typeof r.id !== "string" || !DID_RE.test(r.id)) throw new Error(`principals[${index}].id: invalid DID "${r.id}"`);
+  if (typeof r.public_key !== "string" || !BASE64_RE.test(r.public_key) || r.public_key.length < 40) {
+    throw new Error(`principals[${index}].public_key: invalid (must be base64, ≥40 chars)`);
+  }
+  if (typeof r.operator !== "string" || r.operator.length === 0) throw new Error(`principals[${index}].operator: required`);
+  if (!["agent", "service", "operator"].includes(r.type as string)) throw new Error(`principals[${index}].type: must be agent/service/operator`);
+}
+
 function loadRegistry(): PrincipalRegistryFile {
   if (!existsSync(REGISTRY_PATH)) {
     return { version: 1, principals: [], trusted_hubs: [] };
@@ -63,13 +75,17 @@ function saveRegistry(registry: PrincipalRegistryFile): void {
   console.log(`  registry: ${REGISTRY_PATH}`);
 }
 
+function formatDisplayName(name: string): string {
+  return name.split("-").filter(Boolean).map((w) => w.charAt(0).toUpperCase() + w.slice(1)).join(" ");
+}
+
 export async function generateIdentity(
   name: string,
   operator: string,
   opts: { force?: boolean } = {},
 ): Promise<{ publicKeyB64: string; did: string }> {
   if (!NAMING_RE.test(name)) {
-    throw new Error(`Invalid bot name: "${name}" — must be lowercase alphanumeric + hyphens`);
+    throw new Error(`Invalid bot name: "${name}" — lowercase alphanumeric + hyphens, no trailing/consecutive hyphens`);
   }
 
   ensureKeysDir();
@@ -79,101 +95,114 @@ export async function generateIdentity(
     throw new Error(`Signing key already exists at ${kp}. Use --force to overwrite.`);
   }
 
-  const privateKeyBytes = randomBytes(32);
-  const publicKeyBytes = await getPublicKeyAsync(privateKeyBytes);
+  // Set restrictive umask before writing private key
+  const prevUmask = process.umask(0o077);
+  try {
+    const privateKeyBytes = randomBytes(32);
+    const publicKeyBytes = await getPublicKeyAsync(privateKeyBytes);
 
-  const privateKeyB64 = Buffer.from(privateKeyBytes).toString("base64");
-  const publicKeyB64 = Buffer.from(publicKeyBytes).toString("base64");
+    const privateKeyB64 = Buffer.from(privateKeyBytes).toString("base64");
+    const publicKeyB64 = Buffer.from(publicKeyBytes).toString("base64");
 
-  writeFileSync(kp, privateKeyB64, { mode: 0o600 });
-  chmodSync(kp, 0o600);
+    writeFileSync(kp, privateKeyB64, { mode: 0o600 });
 
-  const did = `did:mf:${name}`;
+    const did = `did:mf:${name}`;
 
-  const registry = loadRegistry();
-  const existing = registry.principals.findIndex((p) => p.id === did);
-  const principal: Principal = {
-    id: did,
-    display_name: name.split("-").map((w) => w[0].toUpperCase() + w.slice(1)).join(" "),
-    operator,
-    public_key: publicKeyB64,
-    type: "agent",
-    created_at: new Date().toISOString(),
-  };
+    const registry = loadRegistry();
+    const existing = registry.principals.findIndex((p) => p.id === did);
+    const principal: Principal = {
+      id: did,
+      display_name: formatDisplayName(name),
+      operator,
+      public_key: publicKeyB64,
+      type: "agent",
+      created_at: new Date().toISOString(),
+    };
 
-  if (existing >= 0) {
-    registry.principals[existing] = principal;
-    console.log(`  updated principal: ${did}`);
-  } else {
-    registry.principals.push(principal);
-    console.log(`  added principal: ${did}`);
+    if (existing >= 0) {
+      registry.principals[existing] = principal;
+      console.log(`  updated principal: ${did}`);
+    } else {
+      registry.principals.push(principal);
+      console.log(`  added principal: ${did}`);
+    }
+    saveRegistry(registry);
+
+    console.log(`  signing key: ${kp} (mode 600)`);
+    console.log(`  public key: ${publicKeyB64}`);
+
+    return { publicKeyB64, did };
+  } finally {
+    process.umask(prevUmask);
   }
-  saveRegistry(registry);
-
-  console.log(`  signing key: ${kp} (mode 600)`);
-  console.log(`  public key: ${publicKeyB64}`);
-
-  return { publicKeyB64, did };
 }
 
 export function exportPrincipals(operator?: string): void {
   const registry = loadRegistry();
-  let principals = registry.principals;
 
+  // Default: export only principals that were locally generated (have a key file)
+  // With -a: filter to that operator
+  let principals = registry.principals;
   if (operator) {
     principals = principals.filter((p) => p.operator === operator);
-    if (principals.length === 0) {
-      console.error(`No principals found for operator "${operator}"`);
-      process.exit(1);
-    }
+  } else {
+    principals = principals.filter((p) => {
+      const name = p.id.replace(/^did:mf:/, "");
+      return existsSync(keyPath(name));
+    });
+  }
+
+  if (principals.length === 0) {
+    console.error(operator
+      ? `No principals found for operator "${operator}"`
+      : "No locally-generated principals found (no matching key files)");
+    process.exitCode = 1;
+    return;
   }
 
   const exportData: PrincipalRegistryFile = {
     version: 1,
     principals,
-    trusted_hubs: registry.trusted_hubs,
+    trusted_hubs: [],
   };
 
-  // Output to stdout for piping
   process.stdout.write(JSON.stringify(exportData, null, 2) + "\n");
 }
 
 export function importPrincipals(filePath: string): void {
   if (!existsSync(filePath)) {
     console.error(`File not found: ${filePath}`);
-    process.exit(1);
+    process.exitCode = 1;
+    return;
   }
 
   let incoming: PrincipalRegistryFile;
   try {
     const raw = JSON.parse(readFileSync(filePath, "utf-8"));
     if (raw.version !== 1 || !Array.isArray(raw.principals)) {
-      throw new Error("Invalid format");
+      throw new Error("expected { version: 1, principals: [...] }");
+    }
+    for (let i = 0; i < raw.principals.length; i++) {
+      validatePrincipal(raw.principals[i], i);
     }
     incoming = raw;
   } catch (err) {
     console.error(`Invalid principals file: ${err instanceof Error ? err.message : String(err)}`);
-    process.exit(1);
+    process.exitCode = 1;
+    return;
   }
 
-  for (const p of incoming.principals) {
-    if (!DID_RE.test(p.id)) {
-      console.error(`Skipping invalid DID: "${p.id}"`);
-      continue;
-    }
+  if (incoming.trusted_hubs?.length > 0) {
+    console.log(`  NOTE: trusted_hubs in import file ignored (security boundary — add manually if needed)`);
   }
 
   const registry = loadRegistry();
   let added = 0;
   let updated = 0;
   let skipped = 0;
+  let rejected = 0;
 
   for (const incoming_p of incoming.principals) {
-    if (!DID_RE.test(incoming_p.id)) {
-      skipped++;
-      continue;
-    }
-
     const existing = registry.principals.findIndex((p) => p.id === incoming_p.id);
     if (existing >= 0) {
       const old = registry.principals[existing];
@@ -181,25 +210,24 @@ export function importPrincipals(filePath: string): void {
         skipped++;
         continue;
       }
-      console.log(`  updated: ${incoming_p.id} (key changed)`);
+      // Key conflict: check operator consistency
+      if (old.operator !== incoming_p.operator) {
+        console.error(`  REJECTED: ${incoming_p.id} — operator mismatch (local: ${old.operator}, import: ${incoming_p.operator}). Manual resolution required.`);
+        rejected++;
+        continue;
+      }
+      console.log(`  UPDATED: ${incoming_p.id} (key rotated, same operator ${old.operator})`);
       registry.principals[existing] = incoming_p;
       updated++;
     } else {
-      console.log(`  added: ${incoming_p.id}`);
+      console.log(`  added: ${incoming_p.id} (operator: ${incoming_p.operator})`);
       registry.principals.push(incoming_p);
       added++;
     }
   }
 
-  // Merge trusted_hubs
-  const hubSet = new Set(registry.trusted_hubs);
-  for (const hub of incoming.trusted_hubs ?? []) {
-    hubSet.add(hub);
-  }
-  registry.trusted_hubs = [...hubSet];
-
   saveRegistry(registry);
-  console.log(`Import complete: ${added} added, ${updated} updated, ${skipped} unchanged`);
+  console.log(`Import complete: ${added} added, ${updated} updated, ${skipped} unchanged, ${rejected} rejected`);
 }
 
 export function listPrincipals(): void {
@@ -212,7 +240,8 @@ export function listPrincipals(): void {
 
   console.log(`Principals (${registry.principals.length}):\n`);
   for (const p of registry.principals) {
-    console.log(`  ${p.id}`);
+    const hasKey = existsSync(keyPath(p.id.replace(/^did:mf:/, "")));
+    console.log(`  ${p.id}${hasKey ? " (local)" : ""}`);
     console.log(`    operator: ${p.operator}`);
     console.log(`    type: ${p.type}`);
     console.log(`    key: ${p.public_key.slice(0, 16)}...`);
--- a/src/commands/identity.ts
+++ b/src/commands/identity.ts
@@ -1,0 +1,223 @@
+/**
+ * arc identity — Myelin signing identity management.
+ *
+ * Generates Ed25519 keypairs for bot signing and manages the
+ * PrincipalRegistry (principals.json) file. Part of grove#320 AAA.
+ */
+
+import { existsSync, readFileSync, writeFileSync, mkdirSync, chmodSync } from "node:fs";
+import { join } from "node:path";
+import { homedir } from "node:os";
+import { getPublicKeyAsync } from "@noble/ed25519";
+import { randomBytes } from "node:crypto";
+
+const CONFIG_BASE = process.env.METAFACTORY_CONFIG_DIR ?? join(homedir(), ".config", "metafactory");
+const KEYS_DIR = join(CONFIG_BASE, "keys");
+const REGISTRY_PATH = join(CONFIG_BASE, "principals.json");
+const DID_RE = /^did:mf:[a-z][a-z0-9._-]+$/;
+const NAMING_RE = /^[a-z][a-z0-9-]*$/;
+
+export interface Principal {
+  id: string;
+  display_name?: string;
+  operator: string;
+  public_key: string;
+  type: "agent" | "service" | "operator";
+  created_at: string;
+  is_hub?: boolean;
+}
+
+export interface PrincipalRegistryFile {
+  version: 1;
+  principals: Principal[];
+  trusted_hubs: string[];
+}
+
+function ensureKeysDir(): void {
+  if (!existsSync(KEYS_DIR)) {
+    mkdirSync(KEYS_DIR, { recursive: true, mode: 0o700 });
+  }
+  chmodSync(KEYS_DIR, 0o700);
+}
+
+function keyPath(name: string): string {
+  return join(KEYS_DIR, `${name}.key`);
+}
+
+function loadRegistry(): PrincipalRegistryFile {
+  if (!existsSync(REGISTRY_PATH)) {
+    return { version: 1, principals: [], trusted_hubs: [] };
+  }
+  const raw = JSON.parse(readFileSync(REGISTRY_PATH, "utf-8"));
+  if (raw.version !== 1 || !Array.isArray(raw.principals)) {
+    throw new Error(`Invalid registry at ${REGISTRY_PATH}: expected version 1 with principals array`);
+  }
+  return raw as PrincipalRegistryFile;
+}
+
+function saveRegistry(registry: PrincipalRegistryFile): void {
+  if (!existsSync(CONFIG_BASE)) {
+    mkdirSync(CONFIG_BASE, { recursive: true });
+  }
+  writeFileSync(REGISTRY_PATH, JSON.stringify(registry, null, 2) + "\n");
+  console.log(`  registry: ${REGISTRY_PATH}`);
+}
+
+export async function generateIdentity(
+  name: string,
+  operator: string,
+  opts: { force?: boolean } = {},
+): Promise<{ publicKeyB64: string; did: string }> {
+  if (!NAMING_RE.test(name)) {
+    throw new Error(`Invalid bot name: "${name}" — must be lowercase alphanumeric + hyphens`);
+  }
+
+  ensureKeysDir();
+  const kp = keyPath(name);
+
+  if (existsSync(kp) && !opts.force) {
+    throw new Error(`Signing key already exists at ${kp}. Use --force to overwrite.`);
+  }
+
+  const privateKeyBytes = randomBytes(32);
+  const publicKeyBytes = await getPublicKeyAsync(privateKeyBytes);
+
+  const privateKeyB64 = Buffer.from(privateKeyBytes).toString("base64");
+  const publicKeyB64 = Buffer.from(publicKeyBytes).toString("base64");
+
+  writeFileSync(kp, privateKeyB64, { mode: 0o600 });
+  chmodSync(kp, 0o600);
+
+  const did = `did:mf:${name}`;
+
+  const registry = loadRegistry();
+  const existing = registry.principals.findIndex((p) => p.id === did);
+  const principal: Principal = {
+    id: did,
+    display_name: name.split("-").map((w) => w[0].toUpperCase() + w.slice(1)).join(" "),
+    operator,
+    public_key: publicKeyB64,
+    type: "agent",
+    created_at: new Date().toISOString(),
+  };
+
+  if (existing >= 0) {
+    registry.principals[existing] = principal;
+    console.log(`  updated principal: ${did}`);
+  } else {
+    registry.principals.push(principal);
+    console.log(`  added principal: ${did}`);
+  }
+  saveRegistry(registry);
+
+  console.log(`  signing key: ${kp} (mode 600)`);
+  console.log(`  public key: ${publicKeyB64}`);
+
+  return { publicKeyB64, did };
+}
+
+export function exportPrincipals(operator?: string): void {
+  const registry = loadRegistry();
+  let principals = registry.principals;
+
+  if (operator) {
+    principals = principals.filter((p) => p.operator === operator);
+    if (principals.length === 0) {
+      console.error(`No principals found for operator "${operator}"`);
+      process.exit(1);
+    }
+  }
+
+  const exportData: PrincipalRegistryFile = {
+    version: 1,
+    principals,
+    trusted_hubs: registry.trusted_hubs,
+  };
+
+  // Output to stdout for piping
+  process.stdout.write(JSON.stringify(exportData, null, 2) + "\n");
+}
+
+export function importPrincipals(filePath: string): void {
+  if (!existsSync(filePath)) {
+    console.error(`File not found: ${filePath}`);
+    process.exit(1);
+  }
+
+  let incoming: PrincipalRegistryFile;
+  try {
+    const raw = JSON.parse(readFileSync(filePath, "utf-8"));
+    if (raw.version !== 1 || !Array.isArray(raw.principals)) {
+      throw new Error("Invalid format");
+    }
+    incoming = raw;
+  } catch (err) {
+    console.error(`Invalid principals file: ${err instanceof Error ? err.message : String(err)}`);
+    process.exit(1);
+  }
+
+  for (const p of incoming.principals) {
+    if (!DID_RE.test(p.id)) {
+      console.error(`Skipping invalid DID: "${p.id}"`);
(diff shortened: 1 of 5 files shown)


src/commands/nats.ts
```diff
@@ -168,6 +169,15 @@ export function addBot(name: string, opts: AddBotOptions): void {
...
   if (opts.sub) console.log(`  subscribe: ${opts.sub}`);
   console.log(`  credentials: ${outPath} (mode 600)`);
+
+  if (opts.withIdentity) {
+    import("./identity").then(({ generateIdentity }) =>
+      generateIdentity(name, account, { force: opts.force }),
+    ).catch((err: Error) => {
+      console.error(`Warning: identity generation failed: ${err.message}`);
+      console.error("NATS credentials were created successfully. Run 'arc identity generate' to retry.");
+    });
```
mellanon: **blocker — fire-and-forget identity generation breaks the headline feature.**

`addBot` is sync (returns `void`). The dynamic `import().then(...).catch(...)` returns a promise that nobody awaits. Three concrete problems:

1. The CLI process can exit before `generateIdentity` completes, leaving the user with NATS creds but no signing key — the exact failure mode the PR claims to eliminate ("one command for NATS creds + identity").
2. If identity generation fails, the catch logs a warning but exit code is still 0 — scripts/CI cannot detect partial failure.
3. Output ordering is racy: the "credentials created" log line and the identity logs interleave non-deterministically.

Fix: make `addBot` async and `await` the import, or call `generateIdentity` synchronously after a single up-front import (or just static-import — `identity.ts` does not import `nats.ts`, so there is no circular-import excuse for the dynamic form here).

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+        continue;
+      }
+      console.log(`  updated: ${incoming_p.id} (key changed)`);
+      registry.principals[existing] = incoming_p;
+      updated++;
+    } else {
+      console.log(`  added: ${incoming_p.id}`);
+      registry.principals.push(incoming_p);
+      added++;
+    }
```
mellanon, line 226: **major — `importPrincipals` accepts attacker-controlled fields with no validation beyond DID shape.**

The only check on each incoming principal is `DID_RE.test(incoming_p.id)`. Everything else is trusted and dropped straight into the registry:

- `public_key` is not validated as base64, not checked for length (Ed25519 public keys are 32 bytes / 44 base64 chars).
- `type` is not constrained to the `"agent" | "service" | "operator"` union — any string passes through.
- `operator` is not validated against `NAMING_RE` or any known account.
- `is_hub: true` smuggled in via the import file is preserved silently — that is precisely the field that grants elevated trust.
- `created_at` is taken from the file rather than stamped on import.

This is a trust-elevation surface: the import flow is described in the PR as "another operator's principals." If that file is tampered with — or socially engineered — the registry silently absorbs it. Validate each principal's shape before merging, reject unknown `type` values, and either strip or reject `is_hub` from imports (an operator should mark a hub themselves, not inherit it).

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+    const existing = registry.principals.findIndex((p) => p.id === incoming_p.id);
+    if (existing >= 0) {
+      const old = registry.principals[existing];
+      if (old.public_key === incoming_p.public_key) {
+        skipped++;
+        continue;
+      }
+      console.log(`  updated: ${incoming_p.id} (key changed)`);
+      registry.principals[existing] = incoming_p;
+      updated++;
```
mellanon, line 221: **major — cross-operator key replacement is silently accepted.**

The update branch compares only `public_key`. There is no check that `old.operator === incoming_p.operator`. So if I import a file from operator B that contains `did:mf:foo` with B's key, and I already have `did:mf:foo` registered under operator A with A's key, B's entry overwrites A's — with a single one-line `updated: did:mf:foo (key changed)` log that an operator running a long import will likely scroll past.

For a credential registry this is the worst kind of footgun: the principal-of-record ownership flips silently. At minimum:

- Reject the update if `old.operator !== incoming_p.operator` unless `--force` (or a new `--allow-rotation`) is passed.
- When a key changes for an existing DID, print old key prefix vs new key prefix so a human can sanity-check.
- Ideally require interactive confirmation for any key-change merge.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+  writeFileSync(kp, privateKeyB64, { mode: 0o600 });
+  chmodSync(kp, 0o600);
+
+  const did = `did:mf:${name}`;
+
+  const registry = loadRegistry();
+  const existing = registry.principals.findIndex((p) => p.id === did);
+  const principal: Principal = {
+    id: did,
+    display_name: name.split("-").map((w) => w[0].toUpperCase() + w.slice(1)).join(" "),
```
mellanon: **major — display_name construction crashes on edge inputs that `NAMING_RE` allows.**

`NAMING_RE = /^[a-z][a-z0-9-]*$/` accepts `foo--bar` (consecutive hyphens) and `foo-` (trailing hyphen). Both produce empty segments after `name.split("-")`, where `w[0]` is `undefined`, and `undefined.toUpperCase()` throws `TypeError: Cannot read properties of undefined`.

Two fixes; pick one:

1. Tighten the regex to forbid those cases, e.g. `/^[a-z][a-z0-9]*(-[a-z0-9]+)*$/`.
2. Filter or guard inside the map: `name.split("-").filter(Boolean).map(w => w[0].toUpperCase() + w.slice(1)).join(" ")`.

Add a test for `arc identity generate foo--bar` so this does not regress.

src/commands/nats.ts
```diff
@@ -168,6 +169,15 @@ export function addBot(name: string, opts: AddBotOptions): void {
   if (opts.pub) console.log(`  publish: ${opts.pub}`);
   if (opts.sub) console.log(`  subscribe: ${opts.sub}`);
   console.log(`  credentials: ${outPath} (mode 600)`);
+
+  if (opts.withIdentity) {
+    import("./identity").then(({ generateIdentity }) =>
```
mellanon: **major — `--force` semantics are conflated.**

`arc nats add-bot foo --with-identity --force` overloads `--force` to mean "overwrite the NATS user **and** overwrite the existing signing key." That's a destructive surprise: an operator retrying after a NATS-only failure silently regenerates the signing key, invalidating every signature ever produced under the old key (until the new public key is re-distributed via `identity export`/`import`).

Two separate concerns, two separate flags. Either:

- Don't pass `force` through to `generateIdentity` from `addBot` — let identity refuse to clobber and print a hint to use `arc identity generate --force` explicitly.
- Or add a separate `--force-identity` flag.

Also: there is no atomicity. NATS-creds-created-but-identity-failed leaves the world in a half-state with no rollback. Either generate identity *first* (cheap, local) and only mint NATS creds after it succeeds, or document the partial-failure recovery path.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  const exportData: PrincipalRegistryFile = {
+    version: 1,
+    principals,
+    trusted_hubs: registry.trusted_hubs,
+  };
+
+  // Output to stdout for piping
+  process.stdout.write(JSON.stringify(exportData, null, 2) + "\n");
+}
```
mellanon, line 170: **major — `arc identity export` with no flag re-exports every operator's principals.**

The default path is `principals = registry.principals` — i.e. *everyone in your registry*, including principals you imported from other operators. The PR description frames this command as "share your principals with another operator." With no `-a/--account` flag, the command instead shares your view of *every operator you have imported from*.

Public keys aren't secrets, so this isn't a credential leak. But it muddies provenance and the operator-of-record. The command-line description says "Filter to one operator's principals" implying default = all-by-design, but the user-facing intent in the PR body says otherwise.

Recommend: require `-a` explicitly (no default), or default to the current operator derived from the same source `arc nats` already uses. Either way, do not silently rebroadcast imported keys.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  const privateKeyBytes = randomBytes(32);
+  const publicKeyBytes = await getPublicKeyAsync(privateKeyBytes);
+
+  const privateKeyB64 = Buffer.from(privateKeyBytes).toString("base64");
+  const publicKeyB64 = Buffer.from(publicKeyBytes).toString("base64");
+
+  writeFileSync(kp, privateKeyB64, { mode: 0o600 });
+  chmodSync(kp, 0o600);
```
mellanon: **major — private-key file write has a small permission window before the chmod.**

```ts
writeFileSync(kp, privateKeyB64, { mode: 0o600 });
chmodSync(kp, 0o600);
```

The `mode` argument to `writeFileSync` is the *creation* mode, masked by the process umask. With a permissive umask the file briefly lands at e.g. `0o644` before the explicit `chmodSync` corrects it. For an Ed25519 *private* key on a multi-user box, that window is non-zero.

Use `fs.openSync(kp, "wx", 0o600)` (or `O_CREAT | O_EXCL | O_WRONLY` with explicit mode) and write to that fd, so the file never exists at any other mode. As a bonus, `wx` will also fail-closed if the file already exists, which is what we want.

While you are here: `saveRegistry` does not constrain `principals.json` mode at all. Public keys aren't secret, so 0o644 is fine — but make it explicit for the same reason (and the matching `mkdirSync(CONFIG_BASE, ...)` should pass `mode: 0o700` to match `KEYS_DIR`).

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+    console.error(`Invalid principals file: ${err instanceof Error ? err.message : String(err)}`);
+    process.exit(1);
+  }
+
+  for (const p of incoming.principals) {
+    if (!DID_RE.test(p.id)) {
+      console.error(`Skipping invalid DID: "${p.id}"`);
+      continue;
+    }
+  }
```
mellanon: **nit — this validation pre-loop is dead.**

```ts
for (const p of incoming.principals) {
  if (!DID_RE.test(p.id)) {
    console.error(`Skipping invalid DID: "${p.id}"`);
    continue;
  }
}
```

The loop has no body apart from the warn-and-continue. Functionally equivalent to a forEach that only logs. The real skip is in the next loop, where `skipped++` happens silently — so invalid DIDs get a one-line warning here, then are silently dropped below.

Either delete this loop and move the `console.error` into the working loop, or delete the `if (!DID_RE.test(...)) { skipped++; continue; }` branch in the second loop and let this pre-pass filter. Right now it's two passes, neither of which is doing the full job.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+  };
+
+  // Output to stdout for piping
+  process.stdout.write(JSON.stringify(exportData, null, 2) + "\n");
+}
+
+export function importPrincipals(filePath: string): void {
+  if (!existsSync(filePath)) {
+    console.error(`File not found: ${filePath}`);
+    process.exit(1);
```
mellanon: **nit — `process.exit(1)` from exported library functions.**

`exportPrincipals`, `importPrincipals` (and the `No principals found` branch in export) call `process.exit(1)` on bad input. That's fine for a CLI action handler but these are *exported* functions — a future caller (say, a test that passes a bad path on purpose, or another command that wants to chain identity work) will have the entire process killed.

Throw, and let `cli.ts` `.action(...)` catch and exit. Same for `generateIdentity`: it already throws (good). Make the three import/export functions match. As a bonus this lets you cover the error paths in tests without hacks.

test/commands/identity.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    exportPrincipals();
+
+    process.stdout.write = origWrite;
+
+    const output = JSON.parse(chunks.join(""));
+    expect(output.version).toBe(1);
+    expect(output.principals).toHaveLength(1);
+    expect(output.principals[0].id).toBe("did:mf:export-bot");
+  });
+});
```
mellanon, line 191: **nit — test coverage skips the security-relevant paths.**

The five tests cover the happy paths. The edges that actually matter for a credential-registry feature are not exercised:

- `importPrincipals` rejecting an invalid file format (truncated JSON, missing `version`, wrong shape).
- `importPrincipals` rejecting an invalid DID (the dead-loop case, see other comment).
- `importPrincipals` updating with a key change (would have caught the cross-operator silent-overwrite issue).
- `importPrincipals` merging `trusted_hubs`.
- `exportPrincipals(operator)` filtering by account, and the "no principals for that operator" exit path.
- `--with-identity` integration on `arc nats add-bot` (the headline feature is untested end-to-end).
- `generateIdentity` with `name = "foo--bar"` (would have caught the display_name crash, see other comment).

Add at least one test per finding once those are fixed — otherwise the regressions come back the next time someone touches this file.

src/commands/nats.ts
```diff
@@ -168,6 +169,15 @@ export function addBot(name: string, opts: AddBotOptions): void {
   if (opts.pub) console.log(`  publish: ${opts.pub}`);
   if (opts.sub) console.log(`  subscribe: ${opts.sub}`);
   console.log(`  credentials: ${outPath} (mode 600)`);
+
+  if (opts.withIdentity) {
```
mellanon, line 174: **nit — dynamic `import("./identity")` is unnecessary.**

`identity.ts` does not import from `nats.ts`. There is no circular-import problem to dodge. A static `import { generateIdentity } from "./identity.js"` at the top of the file is simpler and removes the source of the fire-and-forget bug in the blocker comment above. Once `addBot` is async, this becomes a one-liner: `await generateIdentity(name, account, { force: opts.force });`.

mellanon (changes requested):
Pulled the diff. The shape is right — Ed25519 keypair generation, principals.json registry, export/import lifecycle, plus a `--with-identity` shortcut on `arc nats add-bot`. The execution has gaps that matter for a credential-registry feature.

**Blocker**

1. `addBot` is sync, but `--with-identity` runs `import("./identity").then(...)` and never awaits the promise. The CLI may exit before the signing key exists, partial-failure exits 0, and stdout interleaves non-deterministically. This is the headline feature of the PR. Make `addBot` async and await, and switch to a static import while you are at it.

**Majors**

2. `importPrincipals` validates only the DID shape on incoming entries. `public_key`, `type`, `operator`, `created_at`, and `is_hub` come straight from the file into the registry. `is_hub: true` is the elevation lever; this is a trust-elevation surface, not a parsing nit.
3. Cross-operator key replacement during import is silently accepted (no `old.operator === incoming.operator` check). Principal-of-record ownership flips on a one-line log.
4. `display_name` construction crashes on `foo--bar` and `foo-` inputs that `NAMING_RE` already accepts. Tighten the regex or filter empty segments.
5. `--force` on `arc nats add-bot --with-identity --force` overloads to "regenerate the signing key too." Destructive surprise on retry. Split the concerns or stop passing `force` through.
6. `arc identity export` with no flag exports *every* operator's principals, including ones imported from others. Default = re-broadcast. The PR description frames this as sharing your principals; the default is not that.
7. Private-key file write has a small permission window (umask masks the create-mode before the explicit chmod). Use `fs.openSync(path, "wx", 0o600)` so the file never exists at any other mode. Same file goes for `mkdirSync(CONFIG_BASE)` not constraining mode.

**Nits**

8. The first `for (const p of incoming.principals)` loop in `importPrincipals` has no body other than logging — it's a dead validation pass. The real skip is silent in the next loop.
9. `exportPrincipals` and `importPrincipals` call `process.exit(1)` from exported library code. Throw instead and let the action handler exit.
10. Test coverage is happy-path only. The security-relevant edges (invalid file format, invalid DID, cross-operator key change, `trusted_hubs` merge, `--with-identity` integration, `display_name` edge regex) are unexercised. Several of these findings would have been caught by tests.
11. The `import("./identity")` dynamic form in `nats.ts` is unnecessary — there's no circular dependency.

**What I checked**: CodeQuality, Security, Hardening, Architecture (CLI command-tree shape, identity↔nats coupling), EcosystemCompliance (PrincipalRegistryFile shape vs myelin's spec — looks aligned per the PR body, did not cross-reference myelin source). No Performance lens — this is local fs + one keypair generation per call, not a hot path.

verdict: blockers=1, majors=6, nits=4 — recommend: request-changes

src/commands/nats.ts
```diff
@@ -168,6 +169,15 @@ export function addBot(name: string, opts: AddBotOptions): void {
...
+    import("./identity").then(({ generateIdentity }) =>
+      generateIdentity(name, account, { force: opts.force }),
+    ).catch((err: Error) => {
+      console.error(`Warning: identity generation failed: ${err.message}`);
+      console.error("NATS credentials were created successfully. Run 'arc identity generate' to retry.");
+    });
+  }
```
mellanon, line 176: **[major/architecture]** `addBot` is declared `void` but kicks off fire-and-forget async work when `--with-identity` is set. The CLI action handler in `cli.ts:1191` is not async and does not await this — the dynamic `import("./identity").then(...)` returns a promise that is not propagated.

Two consequences:
1. The exit code stays `0` even when identity generation fails. Scripts running `arc nats add-bot --with-identity` cannot detect partial failures.
2. Output ordering between NATS-success logs and identity logs is by side effect of the event loop, not by contract. Bun happens to drain microtasks before exit; that is not a guarantee to lean on.

Fix: make `addBot` `async`, await `generateIdentity(...)` when `withIdentity`, and have the action handler in `cli.ts` await the result. Errors should propagate as a non-zero exit.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+      console.log(`  updated: ${incoming_p.id} (key changed)`);
+      registry.principals[existing] = incoming_p;
+      updated++;
+    } else {
+      console.log(`  added: ${incoming_p.id}`);
+      registry.principals.push(incoming_p);
+      added++;
+    }
+  }
+
```
mellanon, line 228: **[major/security]** `importPrincipals` silently overwrites the registered public key when an incoming DID matches an existing one. The single `console.log("  updated: ${id} (key changed)")` is trivial to miss in a batch import.

This is the trust boundary of the entire signing scheme: anyone who hands me a `principals.json` can replace the key I associate with `did:mf:andreas-pilot` and I will accept signed messages from their key as if they were from Andreas. Trust-on-first-use is acceptable for phase 1, but silent rotation defeats the point of having identities at all.

Fix options (any of):
- Bail with non-zero exit when an incoming key differs from the existing one, unless `--allow-key-rotation` is passed.
- Group key-change events into the final summary as `WARNING: N key rotations applied` so they cannot be skimmed past.
- Require the user to type `yes` when key conflicts are detected (interactive only).

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+  for (const hub of incoming.trusted_hubs ?? []) {
+    hubSet.add(hub);
+  }
+  registry.trusted_hubs = [...hubSet];
+
+  saveRegistry(registry);
+  console.log(`Import complete: ${added} added, ${updated} updated, ${skipped} unchanged`);
+}
+
+export function listPrincipals(): void {
```
mellanon, line 233: **[major/security]** `importPrincipals` silently merges `trusted_hubs` from the imported file into the local registry. Anyone who sends me a principals export can append entries to my hub trust list without prompting.

Even if `trusted_hubs` is currently unenforced, this is a forward-compat security trap: the moment any code starts gating behavior on `trusted_hubs`, every existing import file becomes a privilege-escalation vector retroactively.

Fix: drop the `trusted_hubs` merge from import. If cross-operator hub sharing is genuinely needed, make it an explicit `arc identity trust-hub <hub>` command so the user has to opt in per hub.

mellanon (changes requested):
See prior review for the full verdict and inline findings. Re-submitting as REQUEST_CHANGES so the PR status reflects it.

verdict: blockers=1, majors=6, nits=4 — recommend: request-changes

src/commands/nats.ts
mellanon: **[warning/quality]** `import("./identity")` is missing the `.js` extension. Every other import in this codebase uses `.js` (see `cli.ts:41`: `from "./commands/identity.js"`).

Bun is permissive enough to resolve this, but strict ESM (Node, or any environment that runs the `bun build` output through Node) will fail with `ERR_MODULE_NOT_FOUND`. This is a portability landmine — the static export compiles fine, the dynamic import compiles fine, then someone runs the bundle on Node and identity provisioning silently breaks under the swallowed catch.

Fix: `import("./identity.js")`.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+  for (const p of incoming.principals) {
+    if (!DID_RE.test(p.id)) {
+      console.error(`Skipping invalid DID: "${p.id}"`);
+      continue;
+    }
+  }
+
+  const registry = loadRegistry();
+  let added = 0;
+  let updated = 0;
```
mellanon, line 201: **[warning/quality]** This first loop has no effect on the import — it logs `Skipping invalid DID` warnings and `continue`s, but tracks no state. The real validation happens in the second loop at line 173, which silently increments `skipped` for the same invalid DIDs.

Net effect: invalid DIDs are warned once (here) and then absorbed into the `unchanged` count of the summary, which is misleading. Either fold the validation into the second loop, or use this loop to fail-fast before any merge work begins (preferred for atomicity — rejecting a bad file partway through a merge leaves the registry in a half-merged state).

src/commands/identity.ts
mellanon: **[warning/quality]** This crashes on names with trailing or consecutive hyphens. `NAMING_RE` (`/^[a-z][a-z0-9-]*$/`) accepts `"jc-"` and `"jc--pilot"`, both of which produce empty segments after `split("-")`. `"".charAt(0)` is `undefined`, and `.toUpperCase()` on `undefined` throws `TypeError`.

Repro: `arc identity generate jc- -a OP_X` → `TypeError: Cannot read properties of undefined`.

Fix: tighten `NAMING_RE` to `/^[a-z][a-z0-9]*(-[a-z0-9]+)*$/`, or guard the split: `name.split("-").filter(Boolean).map(...)`.

src/commands/identity.ts
```diff
@@ -0,0 +1,223 @@
...
+
+function loadRegistry(): PrincipalRegistryFile {
+  if (!existsSync(REGISTRY_PATH)) {
+    return { version: 1, principals: [], trusted_hubs: [] };
+  }
+  const raw = JSON.parse(readFileSync(REGISTRY_PATH, "utf-8"));
+  if (raw.version !== 1 || !Array.isArray(raw.principals)) {
+    throw new Error(`Invalid registry at ${REGISTRY_PATH}: expected version 1 with principals array`);
+  }
+  return raw as PrincipalRegistryFile;
```
mellanon, line 67: **[suggestion/quality]** `loadRegistry` only checks `version === 1` and that `principals` is an array. A `principals.json` containing `{"version": 1, "principals": [{"id": "garbage"}], "trusted_hubs": []}` will pass this gate, then crash later when something reads `p.public_key.slice(0, 16)` (`listPrincipals`) or assumes `p.operator` is a string.

Same problem applies to `importPrincipals` for the incoming file — it validates the DID regex but accepts arbitrary other fields.

Fix: validate each principal's shape (id, public_key, operator, type, created_at all required strings) at load time and at import time. A small helper function reused by both paths is enough.

test/commands/identity.test.ts
```diff
@@ -0,0 +1,120 @@
...
+afterEach(() => {
+  if (testDir && existsSync(testDir)) {
+    rmSync(testDir, { recursive: true, force: true });
+  }
+  delete process.env.METAFACTORY_CONFIG_DIR;
+});
+
+// Re-import each test to pick up env var changes
+async function freshImport() {
+  // Bun caches modules — use a cache-busting query param
```
mellanon: **[suggestion/quality]** This cache-buster is fragile. `CONFIG_BASE`, `KEYS_DIR`, and `REGISTRY_PATH` are evaluated at module top level (identity.ts:14-16), so the only way for `process.env.METAFACTORY_CONFIG_DIR` changes to take effect is to re-execute the module.

Bun does not formally support `?t=...` as a cache key on file URLs — it works today as a side effect of how Bun hashes specifiers, but it is not contractual. A future Bun change could silently start running every test against the same `CONFIG_BASE`, which would be `~/.config/metafactory` of whoever ran the suite. That would corrupt real registries on dev machines.

Fix: read the env var lazily inside each helper instead of at module top level — `function configBase() { return process.env.METAFACTORY_CONFIG_DIR ?? join(homedir(), ".config", "metafactory"); }`. Tests then need no cache-busting.

src/commands/nats.ts
```diff
@@ -168,6 +169,15 @@ export function addBot(name: string, opts: AddBotOptions): void {
   if (opts.pub) console.log(`  publish: ${opts.pub}`);
   if (opts.sub) console.log(`  subscribe: ${opts.sub}`);
   console.log(`  credentials: ${outPath} (mode 600)`);
+
+  if (opts.withIdentity) {
+    import("./identity").then(({ generateIdentity }) =>
+      generateIdentity(name, account, { force: opts.force }),
+    ).catch((err: Error) => {
+      console.error(`Warning: identity generation failed: ${err.message}`);
+      console.error("NATS credentials were created successfully. Run 'arc identity generate' to retry.");
```
mellanon: **[nit/quality]** This catch swallows the failure: prints a warning, then resolves the promise so the process exits `0`. Tied to the `void`-returning `addBot` finding above — once the action handler awaits identity generation, this catch should re-throw (or `process.exitCode = 1`) so scripted callers can detect partial success.

mellanon (changes requested):
## Full Review: the-metafactory/arc#113

### Lens Results
| Lens | Critical | Warning | Suggestion | Nit | Verdict |
|------|----------|---------|------------|-----|---------|
| CodeQuality | 0 | 3 | 2 | 1 | fail |
| Security | 0 | 2 | 0 | 0 | fail |
| Hardening | 0 | 0 | 0 | 0 | n/a (CLI, not API) |
| Architecture | 0 | 1 | 0 | 0 | fail |
| EcosystemCompliance | 0 | 0 | 0 | 0 | pass |
| Performance | 0 | 0 | 0 | 0 | pass |
| Duplication | 0 | 0 | 0 | 0 | pass |

### Key findings
- **[major/security]** `importPrincipals` silently overwrites public keys on DID collision — a single import file can replace the key associated with any DID already in the registry. The trust boundary of the whole signing scheme degrades to "last import wins."
- **[major/security]** `importPrincipals` merges `trusted_hubs` from imported files. The moment any code starts gating on `trusted_hubs`, every existing import becomes a privilege-escalation surface retroactively.
- **[major/architecture]** `addBot` is `void`-returning but kicks off fire-and-forget async identity generation. Exit code stays 0 on failure; output ordering is incidental, not contractual. Action handler must `await`.
- **[warning]** Dynamic `import("./identity")` missing `.js` extension — works on Bun, breaks on strict ESM Node.
- **[warning]** `display_name` derivation crashes on names with trailing/double hyphens that `NAMING_RE` permits.
- **[warning]** Two-pass loop in `importPrincipals` — first loop is dead code, warns about invalid DIDs that the second loop silently absorbs into `unchanged`.
- **[suggestion]** `loadRegistry` only validates shape at the top level, not per principal.
- **[suggestion]** Test cache-busting via `?t=` query param is a Bun implementation detail, not a contract — consider lazy env reads instead.

The shape of the feature is right and the test coverage hits the happy paths. The two security findings are what hold this back: `importPrincipals` is the trust boundary for everything Myelin signs against, and it currently treats key rotation as silent and `trusted_hubs` as transitive trust by file. Either of those blocks merge in my book.

Full review (6 lenses + duplication) — feature shape is right, but `importPrincipals` silently rotates keys and merges trusted_hubs, and `addBot` fire-and-forgets the identity promise.

verdict: blockers=0 majors=3 nits=6 — recommend: request-changes

test/commands/identity.test.ts
mellanon, line 191: **[major/quality]** Tests still skip the paths that the security fixes added. The five tests cover happy-path generate/import/export, but none of the round-3 changes are exercised:

- `importPrincipals` rejecting a cross-operator key conflict (the new `old.operator !== incoming_p.operator` branch at identity.ts:214).
- `importPrincipals` ignoring `trusted_hubs` from the import file (identity.ts:189).
- `validatePrincipal` rejecting malformed input (bad DID, short public_key, missing operator, wrong type).
- `exportPrincipals` filtering to local-only when `-a` is omitted (the keyfile-existence filter at identity.ts:147).
- The umask flow on private-key write.

These are exactly the changes you said this round addressed. A regression in any of them silently re-opens the security finding it was meant to close. Add a test per branch — they're all isolated, no infra needed.

src/commands/identity.ts
```diff
@@ -0,0 +1,252 @@
...
+        skipped++;
+        continue;
+      }
+      // Key conflict: check operator consistency
+      if (old.operator !== incoming_p.operator) {
+        console.error(`  REJECTED: ${incoming_p.id} — operator mismatch (local: ${old.operator}, import: ${incoming_p.operator}). Manual resolution required.`);
+        rejected++;
+        continue;
+      }
+      console.log(`  UPDATED: ${incoming_p.id} (key rotated, same operator ${old.operator})`);
```
mellanon, line 219: **[warning/security]** Same-operator key rotation is still silent — only a `console.log`. The cross-operator REJECT at line 214 closes the loud half of the trust-boundary issue, but the threat that motivated the original blocker isn't fully neutralised: an attacker only needs the victim's DID *and* the operator string (not a secret — operator labels appear in shared docs, ledger entries, and the export files themselves) to land a key rotation. Once landed, every subsequent signature from the impersonated principal verifies.

Two reasonable fixes:
1. Require an explicit `--allow-rotation` flag for any incoming key change, even same-operator. `UPDATED:` becomes opt-in instead of default.
2. Or: require the rotation file be signed by the *current* registered key (cryptographic continuity).

(1) is the one-line fix and gets you the bar this round was supposed to clear. Without one of these, batch imports can still silently rotate keys past a distracted operator.

src/commands/identity.ts
```diff
@@ -0,0 +1,252 @@
...
+import { homedir } from "node:os";
+import { getPublicKeyAsync } from "@noble/ed25519";
+import { randomBytes } from "node:crypto";
+
+const CONFIG_BASE = process.env.METAFACTORY_CONFIG_DIR ?? join(homedir(), ".config", "metafactory");
+const KEYS_DIR = join(CONFIG_BASE, "keys");
+const REGISTRY_PATH = join(CONFIG_BASE, "principals.json");
+const DID_RE = /^did:mf:[a-z][a-z0-9._-]+$/;
+const BASE64_RE = /^[A-Za-z0-9+/]+=*$/;
+const NAMING_RE = /^[a-z](?:[a-z0-9]|-(?=[a-z0-9]))*$/;
```
mellanon, line 19: **[nit/quality]** This `NAMING_RE` is strict — rejects trailing/consecutive hyphens (good, fixes the `formatDisplayName` crash). But `nats.ts:14` still has the *permissive* `/^[a-z][a-z0-9-]*$/`. With `arc nats add-bot foo-- --with-identity`:

1. nats.ts validates and accepts `foo--`.
2. NATS creds get written to disk.
3. Then `generateIdentity` runs, hits the strict regex, throws.

You're left with a half-provisioned bot: creds on disk, no signing key, no principal entry, no rollback. Either tighten `nats.ts` NAMING_RE to match (preferred — share the regex), or validate the name once up front before any disk writes.

mellanon (changes requested):
## Full Review (round 6) — the-metafactory/arc#113

Round-3 fixes mostly landed: cross-operator key conflict now rejects, `trusted_hubs` no longer merges, export defaults to local-only, `addBot` is async + awaited via static import, per-principal validation runs on import, umask wraps the private-key write. The diff is materially safer than round 3.

### Lens Results
| Lens | Critical | Warning | Suggestion | Nit | Verdict |
|------|----------|---------|------------|-----|---------|
| CodeQuality        | 0 | 0 | 0 | 1 | pass (advisory) |
| Security           | 0 | 1 | 0 | 0 | fail |
| Hardening          | 0 | 0 | 0 | 0 | pass |
| Architecture       | 0 | 0 | 0 | 0 | pass |
| EcosystemCompliance| 0 | 0 | 0 | 0 | pass |
| Performance        | 0 | 0 | 0 | 0 | pass |
| Duplication        | 0 | 0 | 0 | 0 | pass |
| **Tests (quality)**| 0 | 0 | 0 | 0 | — |

### Open findings
1. **[major/quality]** Test suite still doesn't cover the security-relevant paths it was supposed to add this round (cross-operator reject, trusted_hubs ignore, validatePrincipal rejection branches, local-only export filter, umask). A regression silently re-opens a closed blocker. *(test/commands/identity.test.ts)*
2. **[warning/security]** Same-operator key rotation is silent — `UPDATED:` log only, no consent. Operator strings aren't secret; the cross-operator REJECT half-closes the trust boundary but doesn't fully close it. Add `--allow-rotation` or require the file be signed by the existing key. *(identity.ts:219)*
3. **[nit/quality]** `NAMING_RE` in `identity.ts` is strict; `nats.ts:14` is permissive. `--with-identity` with an edge-case name leaves a half-provisioned bot (creds written, identity throws). *(identity.ts:19)*

### Overall Verdict: request-changes
The blocker class is gone. The remaining major is the missing tests for the *exact* changes this round delivered — without them, the next refactor can quietly undo the safety. Fix the test gap and either close or justify the same-operator rotation finding, and this is ready.

mellanon (approved):
## Full Review (round 7) — the-metafactory/arc#113

Round-3 closes the test-coverage major and the NAMING_RE sync nit. 11 identity tests on the branch — `cross-operator reject`, `trusted_hubs ignore`, `malformed principal reject`, `invalid file reject`, `local-only export`, plus trailing/consecutive-hyphen rejection. The exact paths I asked be pinned down are now pinned. A regression here will fail CI instead of silently re-opening a closed blocker.

### Lens Results
| Lens | Critical | Warning | Suggestion | Nit | Verdict |
|------|----------|---------|------------|-----|---------|
| CodeQuality        | 0 | 0 | 0 | 0 | pass |
| Security           | 0 | 1 | 0 | 0 | pass (advisory carryover) |
| Hardening          | 0 | 0 | 0 | 0 | pass |
| Architecture       | 0 | 0 | 0 | 0 | pass |
| EcosystemCompliance| 0 | 0 | 0 | 0 | pass |
| Performance        | 0 | 0 | 0 | 0 | pass |
| Duplication        | 0 | 0 | 0 | 0 | pass |
| Tests              | 0 | 0 | 0 | 0 | pass |

### Open advisories (do not block)
1. **[advisory/security]** Same-operator key rotation is still silent — the `UPDATED:` log path on `identity.ts:228` accepts a key flip if the operator string matches. Operator strings aren't secrets, so an attacker who knows your operator name and can hand you an import file can rotate any of your keys with no consent gate. Track as follow-up: `--allow-rotation` flag, or require import files be signed by the existing key. Cross-op REJECT path closes the realistic escalation surface; this is the defense-in-depth half. Not blocking — the realistic attacker doesn't have your operator string AND your trust to import their file. *(carryover from round 6, deliberately deferred)*
2. **[nit]** No direct test asserting the private-key file mode is `0o600` after umask. Single-line config, low regression risk. Optional `statSync(kp).mode & 0o777 === 0o600` would close the loop.

### Overall Verdict: approve

Blocker, all majors, and the one nit I asked for are closed. Tests now cover the security paths I called out, including the cross-operator REJECT path which is the actual trust boundary. Same-operator rotation is a real ask but a defense-in-depth one — landing this and tracking that as a follow-up is the right call.

PR #118, opened 2026-05-11
feat: add ArcPaths + HostAdapter types (#117 phase 1)

jcfischer:
## Summary

Phase 1 of #117 — multi-backend support foundation. Splits the legacy `PaiPaths` concept into two independent concerns:

- **`ArcPaths`** — arc's own host-independent state (configRoot, dbPath, reposDir, cachePath, sourcesPath, secretsDir, runtimeDir, pipelinesDir, actionsDir, shimDir, catalogPath, registryPath)
- **`HostPaths` + `HostAdapter`** — per-backend install dirs and adapter interface (id, detect, paths, supports)

A Claude-Code adapter (`src/lib/hosts/claude-code.ts`) is the default host. New helpers `createArcPaths()` and `getDefaultHost()` build each piece. `createPaths()` is preserved as a deprecated compatibility shim that internally composes ArcPaths + the Claude-Code adapter, and `PaiPaths` is an intersection alias so every existing call site keeps compiling unchanged.

**Zero behavior change in this PR.** All current code paths produce identical output.

## What this enables

Future phases (separate PRs of #117) can now:
- Phase 2: Migrate the 17 `PaiPaths` call sites to use `arc.X` and `host.paths.X` directly.
- Phase 3: Delete `PaiPaths` and `createPaths`.
- Then: Add Codex CLI, Cursor, Continue, Roo Code adapters by implementing `HostAdapter` for each.

## Changes

- `src/types.ts` — Adds `HostId`, `ArcPaths`, `HostPaths`, `HostAdapter`. Marks `PaiPaths` as `@deprecated` (intersection alias).
- `src/lib/hosts/claude-code.ts` — New. Implements the default Claude-Code adapter and `claudeCodePaths()` helper.
- `src/lib/paths.ts` — Adds `createArcPaths()`, `getDefaultHost()`. Refactors `createPaths()` to compose them internally. `migrateConfigIfNeeded` and `findGitRoot` unchanged.
- `test/unit/paths.test.ts` — Adds 9 tests covering `createArcPaths` (4) and `getDefaultHost` (5). Existing 10 tests untouched.

## Design notes

**Why an intersection alias for PaiPaths?** Lets all 17 call sites keep compiling untouched. Each field is marked `@deprecated`, so the TS LSP highlights every Phase-2 migration point automatically. No flag day; PR 2 can migrate file-by-file.

[...]

PR #119, opened 2026-05-11
feat: hostPathFor() dispatch + HostAdapter in install (#117 phase 2)

jcfischer:
## Summary

Phase 2 of #117 — surgical migration of the install hot path to the `HostAdapter` abstraction from Phase 1, plus the type→path mapping Holly asked for in round 3 of #118's review (see https://github.com/the-metafactory/arc/issues/117#issuecomment-4425374042).

## Changes

- **`src/lib/hosts/dispatch.ts`** *(new)* — `hostPathFor(host, type)` returns the host's directory for an artifact type, or `null` for types that don't live inside the host tree:
  - `skill` / `system` → `host.paths.skillsDir`
  - `agent` → `host.paths.agentsDir`
  - `prompt` → `host.paths.promptsDir`
  - `tool` → `host.paths.binDir`
  - `component`, `rules`, `library`, `pipeline`, `action` → `null`
- **`src/lib/artifact-installer.ts`** — `createArtifactSymlinks()` takes a new required `host: HostAdapter` parameter. The skill/agent/prompt/tool dispatch cases use `hostPathFor()` instead of hard-coded `paths.skillsDir` etc. Pipeline still uses `paths.pipelinesDir` because it's host-independent arc state.
- **`src/commands/install.ts`** — `InstallOptions` gains optional `host?: HostAdapter`. `install()` and `installSingleArtifact()` resolve it via `opts.host ?? getDefaultHost(...)`. Existing callers (CLI, tests) keep working unchanged.
- **`test/unit/hosts-dispatch.test.ts`** *(new)* — 10 tests pinning the dispatch table.

## Why `null` instead of throwing in `hostPathFor`

Holly's design note on #117 spelled out the rationale: the mapping should be honest about which types aren't host-installed (rules → consumer repo, pipeline/action → arc state, component → no per-type primary layout, library → meta type). Returning `null` lets callers fall back to the right place rather than inventing a host directory the adapter never promised.

Callers that reach a `null` after asking for what should be a host artifact (skill/agent/prompt/tool) throw a clear `Host \<id\> does not support \<type\> artifacts` error.

## Design choices

**`host?: HostAdapter` optional, not required, on `InstallOptions`.** Making it required would propagate to ~20 callers (CLI + tests). Optional + default = no caller churn for this PR. Phase 3 may tighten this once the rest of the migration is done.

**`getDefaultHost({ root: paths.claudeRoot })`** — explicit pass-through of `paths.claudeRoot` so test isolation works: tests overriding `claudeRoot` automatically get a host pointing at the temp dir without needing to pass `host` explicitly.

**`pipeline` case keeps `paths.pipelinesDir`, not `hostPathFor(host, "pipeline")`.** That's intentional — `hostPathFor` returns `null` for pipelines because they aren't host-installed. The pipeline case continues to use arc state directly. Same logic for `action` (handled in PR 3 when we migrate `paths.actionsDir`).

## Test plan

- [x] `bun test test/unit/hosts-dispatch.test.ts` — 10 pass (new)
- [x] `bun test` full suite — **697/697 pass** (687 prior + 10 new)
- [x] Type-check passes
- [x] Existing `install()` callers untouched
- [x] Holly's design note addressed (issue #117 comment-4425374042)

## What stays for PR 3

Wide mechanical rename of the remaining ~400 `paths.X` accesses to `arc.X` / `host.paths.X`, then deletion of `PaiPaths` and `createPaths`. PR 3 is pure search-and-replace once the abstraction is proven.

## References

- Issue: #117
- Holly's design note: https://github.com/the-metafactory/arc/issues/117#issuecomment-4425374042
- Phase 1 PR: #118 (merged as e2df58f)

### DIFF
--- a/src/lib/artifact-installer.ts
+++ b/src/lib/artifact-installer.ts
@@ -237,8 +237,7 @@ export async function createArtifactSymlinks(opts: {
     }
 
     case "skill":
-    case "system":
-    default: {
+    case "system": {
       // Skills: symlink skill/ subdirectory (or root) to the host's skillsDir.
       const skillsDir = hostPathFor(host, type);
       if (!skillsDir) {
@@ -253,21 +252,33 @@ export async function createArtifactSymlinks(opts: {
         await linkTracked(installDir, skillLinkPath);
       }
 
-      // Create bin symlinks and shims for all CLI entries (skills with CLI).
-      // binDir comes from the same host as the skill install — they ship together.
-      const binDir = hostPathFor(host, "tool");
-      if (!binDir) {
-        throw new Error(`Host ${host.id} does not expose a bin directory for skill CLIs`);
-      }
+      // CLI symlinks + shims only when the skill declares CLI entries.
+      // The binDir lookup must stay inside this guard so a future adapter
+      // that supports skills but exposes no bin directory still installs
+      // pure-content skills cleanly.
       const cliEntries = extractAllCliInfo(manifest);
-      for (const entry of cliEntries) {
-        const binLinkPath = join(binDir, entry.binName);
-        await linkTracked(installDir, binLinkPath);
-      }
       if (cliEntries.length) {
+        const binDir = hostPathFor(host, "tool");
+        if (!binDir) {
+          throw new Error(
+            `Host ${host.id} does not expose a bin directory for skill CLIs`,
+          );
+        }
+        for (const entry of cliEntries) {
+          const binLinkPath = join(binDir, entry.binName);
+          await linkTracked(installDir, binLinkPath);
+        }
         await shimTracked();
       }
       break;
+    }
+
+    default: {
+      // Library is unwound at install.ts:181 before reaching this dispatch;
+      // any other value is a programming bug, not a user-facing error.
+      throw new Error(
+        `Unsupported artifact type "${type as string}" in createArtifactSymlinks`,
+      );
     }
   }
 
--- a/src/lib/artifact-installer.ts
+++ b/src/lib/artifact-installer.ts
@@ -71,13 +71,7 @@ export async function createArtifactSymlinks(opts: {
   type: ArtifactType | "rules" | "system";
   manifest: ArcManifest;
   paths: PaiPaths;
-  /**
-   * Target host adapter. Used to look up per-host install directories via
-   * `hostPathFor(host, type)`. Phase 2 of #117: today it's always the
-   * Claude-Code adapter (so the host paths match what `paths` carries), but
-   * the parameter is here so callers can target Codex/Cursor adapters once
-   * those land.
-   */
+  /** Target host adapter. Defaults to Claude-Code in the caller. */
   host: HostAdapter;
   installDir: string;
   consumerDir?: string;
--- a/src/lib/artifact-installer.ts
+++ b/src/lib/artifact-installer.ts
@@ -2,7 +2,7 @@ import { join, dirname } from "path";
 import { existsSync } from "fs";
 import { mkdir } from "fs/promises";
 import { homedir } from "os";
-import type { ArtifactType, ArcManifest, PaiPaths } from "../types.js";
+import type { ArtifactType, ArcManifest, HostAdapter, PaiPaths } from "../types.js";
 import {
   createSymlink,
   createCliShim,
@@ -11,6 +11,7 @@ import {
   removeCliShim,
 } from "./symlinks.js";
 import { generateRules } from "./rules.js";
+import { hostPathFor } from "./hosts/dispatch.js";
 
 /**
  * Maps an artifact type to its conventional source subdirectory within a cloned repo.
@@ -70,6 +71,14 @@ export async function createArtifactSymlinks(opts: {
   type: ArtifactType | "rules" | "system";
   manifest: ArcManifest;
   paths: PaiPaths;
+  /**
+   * Target host adapter. Used to look up per-host install directories via
+   * `hostPathFor(host, type)`. Phase 2 of #117: today it's always the
+   * Claude-Code adapter (so the host paths match what `paths` carries), but
+   * the parameter is here so callers can target Codex/Cursor adapters once
+   * those land.
+   */
+  host: HostAdapter;
   installDir: string;
   consumerDir?: string;
   quiet?: boolean;
@@ -78,7 +87,7 @@ export async function createArtifactSymlinks(opts: {
   filesMissingSource: Array<{ source: string; target: string }>;
   record: ArtifactSymlinkRecord;
 }> {
-  const { type, manifest, paths, installDir, quiet } = opts;
+  const { type, manifest, paths, host, installDir, quiet } = opts;
   const record: ArtifactSymlinkRecord = { symlinks: [], shims: { dir: paths.shimDir, names: [] } };
 
   // Helper that wraps createSymlink + tracks the target for rollback (#89).
@@ -87,7 +96,7 @@ export async function createArtifactSymlinks(opts: {
     record.symlinks.push(target);
   };
   const shimTracked = async () => {
-    const created = await createCliShim(paths.shimDir, paths.binDir, manifest);
+    const created = await createCliShim(paths.shimDir, host.paths.binDir, manifest);
     record.shims.names.push(...created);
   };
 
@@ -138,16 +147,18 @@ export async function createArtifactSymlinks(opts: {
     }
 
     case "pipeline": {
-      // Pipelines: symlink repo root (or pipeline/ subdirectory) to pipelinesDir
+      // Pipelines: symlink repo root (or pipeline/ subdirectory) to pipelinesDir.
+      // pipelinesDir is arc state (host-independent) — pipelines aren't host-installed.
       const pipelineSourceDir = join(installDir, "pipeline");
       const sourceDir = existsSync(pipelineSourceDir) ? pipelineSourceDir : installDir;
       const pipelineLinkPath = join(paths.pipelinesDir, manifest.name);
       await linkTracked(sourceDir, pipelineLinkPath);
 
-      // If the manifest declares CLI entries, also create shims
+      // If the manifest declares CLI entries, the bin shims still go through
+      // the host (binDir is per-host).
       const cliEntries = extractAllCliInfo(manifest);
       for (const entry of cliEntries) {
-        const binLinkPath = join(paths.binDir, entry.binName);
+        const binLinkPath = join(host.paths.binDir, entry.binName);
         await linkTracked(installDir, binLinkPath);
       }
       if (cliEntries.length) {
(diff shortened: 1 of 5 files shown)


src/lib/hosts/dispatch.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Phase 2 install dispatch (per #117) uses this to ask the host "where does
+ * this go on you?" instead of hard-coding `host.paths.skillsDir` etc. at
+ * every call site. When future adapters land (Codex, Cursor, …) they can
+ * differ on directory naming without changing dispatch code.
+ *
+ * Returning `null` is the right "I cannot install this" signal — callers
+ * should fall back to the arc-state path or the consumer repo, never invent
+ * a host directory the adapter didn't promise.
+ */
+export function hostPathFor(
```
jcfischer, line 36: **[suggestion/architecture]** `supports()` and `hostPathFor() !== null` are not equivalent predicates — and nothing documents this.

`host.supports("component")` → `true`
`hostPathFor(host, "component")` → `null`

Same for `"rules"` and `"library"`. A future developer writing:
```ts
if (host.supports(type)) {
  const dir = hostPathFor(host, type)!;
  // boom — null for component/rules/library
}
```
gets a silent null-pointer.

The two APIs answer different questions: "can the host *work with* this type?" vs "does this type *install into* a host directory?" Today this is safe because the switch cases in `createArtifactSymlinks` only call `hostPathFor` for types that always return non-null. But the contract is implicit — it lives in code-path analysis, not in the type system or docs.

**Fix:** Add a one-line JSDoc on `supports()` (or on `hostPathFor`) clarifying the distinction. Something like: "`supports()` answers whether the host *recognizes* the type; `hostPathFor()` returns `null` when recognition doesn't imply a host-owned directory (component, rules, library)."

src/lib/artifact-installer.ts
```diff
@@ -163,15 +174,19 @@ export async function createArtifactSymlinks(opts: {
     }
 
     case "tool": {
-      // Tools: symlink repo root to binDir for each CLI entry
+      // Tools: symlink repo root to the host's binDir for each CLI entry.
+      const binDir = hostPathFor(host, "tool");
+      if (!binDir) {
+        throw new Error(`Host ${host.id} does not support tool artifacts`);
```
jcfischer, line 174: **[suggestion/quality]** No test exercises the null-guard throws.

With only the claude-code adapter (which returns non-null for skill/agent/prompt/tool), the five `if (!dir) throw` branches on lines 179-180, 200-201, 221-222, 244-245, 259-260 are dead code — they compile but never fire.

When a future adapter arrives that doesn't support some type (say Codex doesn't support `agent`), these throws will be the error surface and they'll land untested.

**Fix:** Add one test with a minimal stub adapter (`{ id: "stub", paths: {...}, detect: () => false, supports: () => false }`) that hands back real paths but where `hostPathFor` returns null for a given type. Call `createArtifactSymlinks` with type `"agent"` on that adapter, assert it throws `"Host stub does not support agent artifacts"`. One test covers the pattern for all five branches.

src/lib/artifact-installer.ts
```diff
@@ -180,56 +195,73 @@ export async function createArtifactSymlinks(opts: {
...
       } else {
         await linkTracked(installDir, skillLinkPath);
       }
 
-      // Create bin symlinks and shims for all CLI entries (skills with CLI)
+      // Create bin symlinks and shims for all CLI entries (skills with CLI).
+      // binDir comes from the same host as the skill install — they ship together.
+      const binDir = hostPathFor(host, "tool");
+      if (!binDir) {
+        throw new Error(`Host ${host.id} does not expose a bin directory for skill CLIs`);
```
jcfischer: **[suggestion/duplication]** The null-guard-then-throw pattern repeats 5 times.

```ts
const dir = hostPathFor(host, "X");
if (!dir) { throw new Error(`Host ${host.id} does not support X artifacts`); }
```

Lines 178-180, 199-201, 221-222, 243-245, 258-260. Four of the five have identical structure (the fifth says "does not expose a bin directory for skill CLIs" — slightly different wording, same intent).

Not blocking — 3 lines each isn't egregious. But if you're touching this file in Phase 3 anyway, a `requireHostDir(host: HostAdapter, type: ArtifactType | "system"): string` helper that throws on null would collapse each site to one line and centralize the error format.

src/lib/artifact-installer.ts
```diff
@@ -138,16 +147,18 @@ export async function createArtifactSymlinks(opts: {
...
       const pipelineLinkPath = join(paths.pipelinesDir, manifest.name);
       await linkTracked(sourceDir, pipelineLinkPath);
 
-      // If the manifest declares CLI entries, also create shims
+      // If the manifest declares CLI entries, the bin shims still go through
+      // the host (binDir is per-host).
       const cliEntries = extractAllCliInfo(manifest);
       for (const entry of cliEntries) {
-        const binLinkPath = join(paths.binDir, entry.binName);
+        const binLinkPath = join(host.paths.binDir, entry.binName);
```
jcfischer, line 155: **[nit/architecture]** Pipeline case accesses `host.paths.binDir` directly (line 161) while the `tool` and `skill/default` cases go through `hostPathFor(host, "tool")`. Both resolve to the same value today.

The PR description explains this is intentional — pipeline itself goes to arc state (`paths.pipelinesDir`), only its CLI shims need a host dir. `hostPathFor(host, "pipeline")` correctly returns null, so using the dispatch table for the shim's bin target would mean `hostPathFor(host, "tool")` which is semantically odd (it's not a tool, it's a pipeline's CLI entry).

Just noting the asymmetry for future readers. Not suggesting a change.

jcfischer (commented):
Full review (6 lenses + duplication) — clean dispatch table design, surgical scope, zero behavior change, 697/697 tests. Four suggestions: document supports() vs hostPathFor() semantic gap, add stub-adapter test for null-guard throws, update CLAUDE.md architecture section, consider extracting requireHostDir helper in Phase 3.

verdict: blockers=0 majors=0 nits=5 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/lib/artifact-installer.ts
```diff
@@ -180,56 +195,73 @@ export async function createArtifactSymlinks(opts: {
...
         await linkTracked(skillSourceDir, skillLinkPath);
       } else {
         await linkTracked(installDir, skillLinkPath);
       }
 
-      // Create bin symlinks and shims for all CLI entries (skills with CLI)
+      // Create bin symlinks and shims for all CLI entries (skills with CLI).
+      // binDir comes from the same host as the skill install — they ship together.
+      const binDir = hostPathFor(host, "tool");
+      if (!binDir) {
```
jcfischer: **[warning/architecture]** `hostPathFor(host, "tool")` null-guard fires before `extractAllCliInfo()`. If a future host adapter returns null for `"tool"` (no bin directory), every skill install fails — even skills with zero CLI entries. The guard should move inside the `if (cliEntries.length)` block, consistent with the PR's own null-return-means-fallback design.

**Fix:** Move lines 258-260 inside the `if (cliEntries.length)` block (after line 266), and add the null check only when there are actual CLI entries to install.

src/lib/artifact-installer.ts
```diff
@@ -180,56 +195,73 @@ export async function createArtifactSymlinks(opts: {
...
         // Fallback: symlink directory if .md file not found by convention name
-        await linkTracked(sourceDir, join(paths.promptsDir, manifest.name));
+        await linkTracked(sourceDir, join(promptsDir, manifest.name));
       }
       break;
     }
 
     case "skill":
     case "system":
     default: {
```
jcfischer: **[suggestion/quality]** The `default` case catches `"library"` type. Libraries never reach `createArtifactSymlinks` today (caught by `install()` upstream), but if called directly the error message is `"Host does not support skill artifacts"` — misleading for a library type. Consider an explicit `case "library": throw new Error('library is a meta type; install contained artifacts individually')` above the default.

jcfischer (commented):
Full review (6 lenses + duplication) — clean dispatch abstraction with one latent null-guard ordering bug in the skill CLI path that violates the PR's own fallback contract.

verdict: blockers=0 majors=0 warnings=1 suggestions=3 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

jcfischer (commented):
Full review cycle 2 (6 lenses + duplication) — JSDoc and stub-test fixes from cycle 1 are solid. Two carryover findings remain (binDir guard position, library in default case) but neither fires on any current code path.

verdict: blockers=0 majors=0 nits=2 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

jcfischer (commented):
Full review (6 lenses + duplication) — clean migration of install hot path to HostAdapter. Cycle-1 findings addressed. One latent multi-host bug in the dependency recursive call (suggestion), two nits.

verdict: blockers=0 majors=0 nits=3 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/lib/artifact-installer.ts
```diff
@@ -138,16 +147,18 @@ export async function createArtifactSymlinks(opts: {
...
 
-      // If the manifest declares CLI entries, also create shims
+      // If the manifest declares CLI entries, the bin shims still go through
+      // the host (binDir is per-host).
       const cliEntries = extractAllCliInfo(manifest);
       for (const entry of cliEntries) {
-        const binLinkPath = join(paths.binDir, entry.binName);
+        const binLinkPath = join(host.paths.binDir, entry.binName);
         await linkTracked(installDir, binLinkPath);
       }
```
jcfischer, line 157: **[suggestion/architecture]** Pipeline case bypasses the binDir guard pattern.

The `tool` case (line 178) and `skill/system` case (line 261) both use `hostPathFor(host, "tool")` with a null/falsy guard before writing CLI symlinks into binDir. The pipeline case (line 161) uses `host.paths.binDir` directly — no guard.

Today this is safe (only the Claude-Code adapter exists, and it always sets a real binDir). But a future adapter with `binDir: ""` would get a clean throw from tool/skill installs and a silent relative-path symlink from pipeline installs. Your own test (`skills without CLI entries install cleanly even when host has no binDir`) creates exactly this adapter shape.

Fix: wrap the pipeline CLI block the same way skill/system does:
```typescript
if (cliEntries.length) {
  const binDir = hostPathFor(host, "tool");
  if (!binDir) {
    throw new Error(`Host ${host.id} does not expose a bin directory for pipeline CLIs`);
  }
  for (const entry of cliEntries) {
    const binLinkPath = join(binDir, entry.binName);
    await linkTracked(installDir, binLinkPath);
  }
  await shimTracked();
}
```

jcfischer (commented):
Full review (6 lenses + duplication) — cycle 3 re-review, both carryovers fixed and tested, zero new findings across all lenses.

verdict: blockers=0 majors=0 nits=0 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

src/lib/artifact-installer.ts
```diff
@@ -70,6 +71,14 @@ export async function createArtifactSymlinks(opts: {
...
   manifest: ArcManifest;
   paths: PaiPaths;
+  /**
+   * Target host adapter. Used to look up per-host install directories via
+   * `hostPathFor(host, type)`. Phase 2 of #117: today it's always the
+   * Claude-Code adapter (so the host paths match what `paths` carries), but
+   * the parameter is here so callers can target Codex/Cursor adapters once
+   * those land.
+   */
+  host: HostAdapter;
```
jcfischer, line 75: **[nit/quality]** This JSDoc is 7 lines for a 1-sentence concept ("defaults to Claude-Code"). The phase/PR references ("… Phase 2 of #117 … once Phase 2's host detection lands") will rot as the migration progresses. Consider trimming to the one sentence that matters:

```typescript
/** Target host adapter. Defaults to Claude-Code when omitted. */
```

The design rationale is well-covered in the PR description and issue #117.

jcfischer (commented):
Full review (6 lenses + duplication) — clean implementation of Phase 2 dispatch with one guard-consistency gap in the pipeline case and minor doc hygiene.

verdict: blockers=0 majors=0 nits=4 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #121, opened 2026-05-12
feat(hosts): CortexHostAdapter (#117 phase 2)

jcfischer:
## Summary

Ship `createCortexHost()` adapter so `arc install <bot> --target cortex` can route agent artifacts into a running cortex install. Lands alongside the existing claude-code adapter as the second `HostAdapter` implementation, matching the Phase 1 contract from #118 (`id` / `detect()` / `paths` / `supports()`).

Unblocks cortex F-5 (Phase A.3 of cortex#58 — `docs/design-arc-agent-bots.md` §6.2). Closes cortex's arc-installable-bots dependency on arc#117 Phase 1.

## What changed

- **`src/types.ts`** — `HostId` widens to `"claude-code" | "cortex"`; new `CortexPaths` interface (`personasDir`, `credsDir`) and `CortexHostPaths = HostPaths & CortexPaths` alias.
- **`src/lib/hosts/cortex.ts`** *(new, ~95 LOC)* — `cortexPaths(opts)` + `createCortexHost(opts)` factory. Mirrors `claude-code.ts` style. Returns `HostAdapter & { paths: CortexHostPaths }` so cortex-aware callers see the extension while generic dispatch still works off plain `HostPaths`.
- **`test/unit/hosts-cortex.test.ts`** *(new, 14 tests)* — paths/detect/supports/integration coverage. Uses an injectable `cortexOnPath` hook for deterministic `detect()` tests.

Cortex hosts agent identity fragments + persona files + per-agent NATS creds. It is **not** a skills/prompts/tools host — `supports("agent")` is the only `true` arm; non-agent dirs are empty strings so dispatch's `if (!dir)` guard catches mis-routed installs.

## Test plan

- [x] `bun test test/unit/hosts-cortex.test.ts` — 14 pass / 0 fail
- [x] `bun test` (full suite) — 718 pass / 0 fail (no regressions)
- [x] `bunx tsc --noEmit` — clean
- [x] `hostPathFor(cortexHost, "agent")` integration test confirms cortex slots into existing dispatch with zero changes to `dispatch.ts` / `artifact-installer.ts`

## Out of scope (explicitly deferred)

- **`installArtifact` / `removeArtifact` adapter methods.** arc#117 Phase 1 deferred them to the `hostPathFor` / `requireHostDir` dispatch helpers. Cortex's spec §6.2 wrote them assuming an aspirational interface; this PR ships against the actual Phase 1 interface and defers per-adapter install hooks to a follow-up when arc adopts them upstream.
- **Cortex post-install side effects** (`cortex agents reload`, `cortex creds issue <id>`). Belong on the cortex side once the adapter API exposes a post-install callback. Filed conceptually for cortex F-5 follow-up.
- **NATS health probe in `detect()`.** The §6.2 design mentions `local.{org}.cortex.health` reachability. Sync `detect()` on every CLI invocation can't realistically open a NATS connection; v1 uses config-file presence + `cortex` binary on PATH (parity with claude-code adapter's dir-existence detection).
- **`hooksFormat: "none"`** on `HostPaths`. Spec §6.2 mentions this field but arc#117 Phase 1 `HostPaths` does not declare it. Not adding without an upstream arc decision.
- **`personasDir` / `credsDir` promotion to base `HostPaths`.** Kept as cortex-only extension per spec §6.2 v1 preference — no other host has the same concepts.
- **`getDefaultHost()` change.** Default stays claude-code. Adapter-detection selection across multiple installed hosts is arc#117 Phase 2 broader scope; out of scope here.

## Related

- arc#117 (umbrella — multi-backend host adapters)
- arc#118 (Phase 1 — `ArcPaths` + `HostAdapter` types)
- arc#119 (Phase 2 — `hostPathFor` dispatch + install integration)
- arc#120 (Phase 3a — `requireHostDir` helper + `arc/host` on `TestEnv`)
- cortex `docs/design-arc-agent-bots.md` §6.2 (CortexHostAdapter spec)
- cortex#58 (umbrella — arc-installable sub-bots)

### DIFF
--- a/src/lib/hosts/cortex.ts
+++ b/src/lib/hosts/cortex.ts
@@ -1,0 +1,92 @@
+import { join } from "path";
+import { existsSync } from "fs";
+import { homedir } from "os";
+import { execSync } from "child_process";
+import type {
+  ArtifactType,
+  CortexHostPaths,
+  HostAdapter,
+} from "../../types.js";
+
+/**
+ * Cortex host adapter.
+ *
+ * Hosts agent identity fragments (`agents.d/<id>.yaml`), persona files
+ * (`~/.config/cortex/personas/<id>.md`), and per-agent NATS creds files
+ * (`~/.config/nats/creds/<id>.creds`). Cortex does NOT host skills, prompts,
+ * or tools — those belong to a claude-code or codex host.
+ *
+ * Detection: a cortex install is recognized when `~/.config/cortex/cortex.yaml`
+ * exists OR a `cortex` binary is on PATH. The NATS health probe envisioned in
+ * `docs/design-arc-agent-bots.md` §6.2 is deferred — too heavy for sync
+ * `detect()`, flaky in CI, and config-file presence already covers v1.
+ *
+ * See cortex `docs/design-arc-agent-bots.md` §6.2 for the full design
+ * rationale and the post-install side effects (`cortex agents reload`,
+ * `cortex creds issue`) that land in a follow-up once arc exposes per-adapter
+ * install hooks.
+ */
+
+export interface CortexHostOptions {
+  /** Cortex config root (default: ~/.config/cortex). */
+  configRoot?: string;
+  /**
+   * Directory where the cortex daemon writes per-agent NATS creds
+   * (default: ~/.config/nats/creds). Kept separate from `configRoot`
+   * because NATS clients expect creds at the NATS-conventional location.
+   */
+  credsRoot?: string;
+  /**
+   * Override the cortex-binary-on-PATH check used by `detect()`. Useful in
+   * tests that want to assert config-file-only detection without depending
+   * on the dev host's actual PATH.
+   */
+  cortexOnPath?: () => boolean;
+}
+
+/** Build cortex host paths rooted at the given config root + creds root. */
+export function cortexPaths(opts?: CortexHostOptions): CortexHostPaths {
+  const configRoot = opts?.configRoot ?? join(homedir(), ".config", "cortex");
+  const credsRoot =
+    opts?.credsRoot ?? join(homedir(), ".config", "nats", "creds");
+  return {
+    root: configRoot,
+    // Cortex is not a skills host — skillsDir stays empty so a caller that
+    // bridges `supports()` to a path lookup gets an obviously-broken value
+    // instead of a plausibly-wrong one. `supports("skill") === false` is the
+    // truthful gate; `hostPathFor(host, "skill")` returning `""` here is
+    // defensive belt-and-suspenders.
+    skillsDir: "",
+    agentsDir: join(configRoot, "agents.d"),
+    // Cortex doesn't host slash commands or operator binaries; both are empty
+    // for the same reason as skillsDir.
+    promptsDir: "",
+    binDir: "",
+    settingsPath: join(configRoot, "cortex.yaml"),
+    // Cortex-only extensions:
+    personasDir: join(configRoot, "personas"),
+    credsDir: credsRoot,
+  };
+}
+
+function defaultCortexOnPath(): boolean {
+  try {
+    execSync("command -v cortex", { stdio: "ignore" });
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export function createCortexHost(opts?: CortexHostOptions): HostAdapter & {
+  paths: CortexHostPaths;
+} {
+  const paths = cortexPaths(opts);
+  const cortexOnPath = opts?.cortexOnPath ?? defaultCortexOnPath;
+  return {
+    id: "cortex",
+    paths,
+    detect: () => existsSync(paths.settingsPath) || cortexOnPath(),
+    supports: (type: ArtifactType) => type === "agent",
+  };
+}
--- a/src/lib/hosts/cortex.ts
+++ b/src/lib/hosts/cortex.ts
@@ -1,7 +1,6 @@
 import { join } from "path";
 import { existsSync } from "fs";
 import { homedir } from "os";
-import { execSync } from "child_process";
 import type {
   ArtifactType,
   CortexHostPaths,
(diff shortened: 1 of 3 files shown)


src/lib/hosts/cortex.ts
```diff
@@ -0,0 +1,92 @@
...
+    // Cortex is not a skills host — skillsDir stays empty so a caller that
+    // bridges `supports()` to a path lookup gets an obviously-broken value
+    // instead of a plausibly-wrong one. `supports("skill") === false` is the
+    // truthful gate; `hostPathFor(host, "skill")` returning `""` here is
+    // defensive belt-and-suspenders.
+    skillsDir: "",
+    agentsDir: join(configRoot, "agents.d"),
+    // Cortex doesn't host slash commands or operator binaries; both are empty
+    // for the same reason as skillsDir.
+    promptsDir: "",
```
mellanon, line 61: **[major/quality]** `defaultCortexOnPath()` uses `execSync("command -v cortex")`, which spawns `/bin/sh -c "command -v cortex"`. `command` is a POSIX shell builtin and not available on Windows (where `/bin/sh` does not exist). claude-code's `detect()` is a pure `existsSync(root)` and is cross-platform; cortex's now has a Unix-only fallback path.

The PR body draws parity between "binary on PATH" and claude-code's "dir-existence detection" — they're not equivalent in cross-platform terms. Two ways out: (a) drop the PATH probe and rely on `cortex.yaml` presence only (config-file detection is already documented as sufficient for v1 in the JSDoc), or (b) use a platform-aware `which` (e.g. `process.env.PATH` walk + `which.sync` from a tiny dep, or `where` on win32). Option (a) is simpler and the JSDoc already says "config-file presence already covers v1" — the binary-on-PATH branch may not be carrying its weight.

src/lib/hosts/cortex.ts
```diff
@@ -0,0 +1,92 @@
...
+    binDir: "",
+    settingsPath: join(configRoot, "cortex.yaml"),
+    // Cortex-only extensions:
+    personasDir: join(configRoot, "personas"),
+    credsDir: credsRoot,
+  };
+}
```
mellanon, line 68: **[nit/performance]** `detect()` on this adapter is heavier than claude-code's by an order of magnitude — claude-code does a single `existsSync`, cortex falls through to a child-process spawn (`execSync` of `/bin/sh -c "command -v cortex"`) whenever the config file is missing.

For a single CLI invocation that's fine. If dispatch ever loops over installed hosts to find one that supports an artifact type (likely shape of the Phase 2 multi-host selection in #117), this becomes N shell spawns per invocation, all to answer a question whose answer can't change within a process lifetime. Worth either (a) memoizing `defaultCortexOnPath` at module scope, or (b) just dropping the PATH probe (same point as the major comment — if config-file presence is the documented v1 contract, the binary probe is doing double duty).

Linking this to the major because the fix for both is the same.

mellanon (changes requested):
## Echo — Full Review (arc#121)

Adversarial pass. 6 lenses + duplication. Diff is small, well-scoped, well-documented.

### Lenses applied
| Lens | Findings | Verdict |
|------|---------:|---------|
| CodeQuality | 1 nit | advisory |
| Security | 0 | pass |
| Hardening | 0 (N/A — library code, no API surface) | pass |
| Architecture | 0 | pass |
| EcosystemCompliance | 0 | pass |
| Performance | 1 nit (rolled into the major) | advisory |
| Duplication | 0 | pass |

### Surface area covered
- `src/lib/hosts/cortex.ts` — full read, line by line
- `src/types.ts` — `HostId` widening, `CortexPaths`/`CortexHostPaths` shape
- `test/unit/hosts-cortex.test.ts` — 14 tests
- Cross-checked against `claude-code.ts` (parity), `dispatch.ts` (integration contract), `paths.ts` and `install.ts` (call sites — no exhaustive switches on `HostId`, so the union widening is safe).

### Key findings
1. **[major/quality]** `defaultCortexOnPath()` shells out via `execSync("command -v cortex")` — `command` is a POSIX builtin; this branch is Unix-only. claude-code's `existsSync` is cross-platform. PR body claims parity between the two detect strategies; not equivalent on Windows. Two clean fixes (drop PATH probe / use a platform-aware which); JSDoc already says config-file presence is sufficient for v1, so option (a) costs nothing. Inline comment posted.
2. **[nit/perf]** Same code — `execSync` makes `detect()` an order of magnitude heavier than claude-code's. Fine for one CLI invocation, less fine if Phase 2 dispatch ever loops over installed hosts. Same fix as the major. Inline comment posted.

### What I did NOT flag
- Out-of-scope items in PR body (install/remove hooks, `cortex agents reload`, NATS health probe, `getDefaultHost` change, `hooksFormat`, base-`HostPaths` promotion) — all pre-flagged by Luna in the PR body with sound reasoning. No second-guessing.
- Empty `skillsDir`/`promptsDir`/`binDir` strings: documented as defensive belt-and-suspenders; `supports()` is the truthful gate; the alternative (`skillsDir?: string` on base `HostPaths`) is a much bigger blast radius. Trade-off is correct.
- `HostId` union widening: grepped repo for exhaustive switches; none. Safe.
- `cortexPaths` vs `claudeCodePaths`: structurally similar but semantically divergent (different fields populated). Not duplication — extracting would be a false hoist.

### Diff hygiene
- 303 / -1, three files. 14 tests, 718/718 suite green per PR body. `tsc --noEmit` clean per PR body.
- JSDoc quality is unusually high — every non-obvious choice has a comment explaining it.

verdict: blockers=0, majors=1, nits=1 — recommend: request-changes

mellanon (approved):
## Echo — Re-Review (arc#121, round 2)

Round 1 found one major and one nit, both on the same code path (`execSync("command -v cortex")` PATH probe). Both resolved against `b11cebd`.

### Round 1 → round 2 resolution

| Round 1 finding | Status | Evidence in `b11cebd` |
|---|---|---|
| **[major/quality]** `execSync("command -v cortex")` PATH probe is POSIX-only — not cross-platform parity with claude-code's `existsSync` | **resolved** | `child_process` import gone; `defaultCortexOnPath()` deleted; `cortexOnPath` option deleted; `detect()` is now `existsSync(paths.settingsPath)` — direct mirror of `claude-code.ts`. JSDoc on `cortex.ts:11-23` explicitly explains why the binary probe was rejected (POSIX `/bin/sh` assumption, platform-specific `which`/`where` split, heavier than `existsSync`). |
| **[nit/performance]** `execSync` makes `detect()` an order of magnitude heavier than claude-code's; compounds under multi-host dispatch loops | **resolved** | Same fix — one `existsSync` call, no shell spawn. Comparable to claude-code's cost. |
| Pinning suggestion ("future re-introduction must update the assertion explicitly") | **honored** | New test at `test/unit/hosts-cortex.test.ts:113-130` — "detect() ignores cortex binary on PATH — config file is the sole signal" — references arc#121 by name and pins the contract. |

### Lenses applied this round (6 + duplication)

| Lens | Findings (new) | Verdict |
|------|---------------:|---------|
| CodeQuality | 0 | pass |
| Security | 0 (N/A — library code, no I/O surface) | pass |
| Hardening | 0 (N/A — no API/middleware/Worker) | pass |
| Architecture | 0 | pass |
| EcosystemCompliance | 0 | pass |
| Performance | 0 | pass |
| Duplication | 0 | pass |

### Surface area covered
- `src/lib/hosts/cortex.ts` — full read, line by line. Verified `detect()` mirrors claude-code's `existsSync` exactly; verified no `child_process` import; verified `cortexOnPath` option gone from the public interface.
- `src/types.ts` — `HostId` widening + `CortexPaths`/`CortexHostPaths` shape unchanged from round 1, still safe (no exhaustive switches on `HostId` in the repo).
- `test/unit/hosts-cortex.test.ts` — all 14 tests reviewed. The new pinning test (lines 113-130) does exactly what was asked of it; the contract is now explicitly enforced.
- Cross-checked against `claude-code.ts` for parity (now genuine parity, both use `existsSync`).

### Diff hygiene
- +275 / -1, three files. 14 unit tests, 718/718 full suite green per PR body. `tsc --noEmit` clean per PR body. Conventional commits.
- JSDoc still high-quality — every non-obvious choice has a comment explaining it. The "Detection" paragraph at `cortex.ts:11-23` is now a fuller explanation of why the binary probe was rejected, which is the right outcome from a re-review cycle.

### Verdict
Clean re-review. Both round-1 findings addressed at the recommended fix (option a — drop the PATH probe, config-file-only). Pinning test means future contributors can't silently re-introduce the regression. No new findings across any lens. Recommend merge.

verdict: blockers=0, majors=0, nits=0 — recommend: merge

PR #132, opened 2026-05-12
fix(nats): server-side revoke + JWT push before delete (closes #130)

mellanon:
## What

Closes **arc#130** — real security gap. `arc nats remove-bot` and `arc nats reissue-bot` previously ran only `nsc delete user`, removing the user from the local NSC tree without pushing the revocation to the NATS server's $SYS account. The user JWT stayed valid on the bus until natural expiry. Any leaked copy of the `.creds` file (including the `.bak` file `reissueBot` itself writes) continued to authenticate.

This PR pushes the actual revocation before the local delete.

## Fix shape

New helpers in `src/commands/nats.ts`:
- **`getUserPubKey(account, name)`** — runs `nsc describe user -J`, parses JSON, validates `sub` is a U-prefixed NKey. Keyed by durable identity (pubkey), not the local name.
- **`revokeAndPushUser(account, name)`** — calls `nsc revocations add-user -u <pubkey>` then `nsc push -a <account>`. Throws with a "STILL VALID on the bus" message on either failure.

Updated commands:
- **`removeBot`**: calls `revokeAndPushUser` BEFORE `nsc delete user`. On failure, prints the error and `process.exit(1)` — never reaches the delete (half-done revoke is worse than failed revoke).
- **`reissueBot`**: revokes the OLD pubkey + pushes BEFORE `delete user` / `add user` / `generate creds`. CRITICAL-error message rewritten to reflect new server-side guarantee (old creds are revoked, backup is forensics-only).

## Pre-flight decisions

| Question | Issue body suggestion | What shipped | Rationale |
|---|---|---|---|
| Pubkey extraction | `nsc describe user -F sub` | `-J` + `JSON.parse(...).sub` + U-prefix validation | `-F sub` semantics not verifiable; `-J` returns documented full JWT claim |
| Revocation verb | `nsc revocations add-user` | Same, with `-u <pubkey>` form | Revocation map keyed by durable identity, not local name |
| Push surface | `nsc push -a <account>` | Same | Account-scoped push to operator's configured JWT server |

## Surprises in the existing code

- `nsc revocations add-user` accepts `-n <name>` pre-delete too — but the pubkey-keyed form (`-u`) is correct per the issue's intent: revocation entries must survive local user-record deletion.
- The existing `nsc` helper inlined `Bun.spawnSync` — refactored to an `NscRunner` interface (production behaviour unchanged; allows test-time injection). Same pattern for `ensureNscInstalled()` via parallel `__setNscInstallCheckForTests` hook.

## Tests (+7 new, all mock-based)

`test/commands/nats-revoke.test.ts`:
1. `removeBot` invokes `revocations add-user` + `push` BEFORE `delete user` (order asserted)
2. `removeBot` aborts when `push` fails — `delete user` is NEVER called, exit code 1, "STILL VALID" message surfaced
3. `removeBot` aborts when `revocations add-user` fails — neither `push` nor `delete user` is called
4. `reissueBot` revokes OLD pubkey + pushes BEFORE `delete user` / `add user` / `generate creds`
5. `reissueBot` aborts on push failure — none of delete/add/generate fire
6. `getUserPubKey` surfaces clean error on malformed JSON
7. `getUserPubKey` rejects a `sub` that is not a U-prefixed NKey

Writes are isolated to `mkdtempSync(tmpdir(), "arc-a130-")`; unit tests never touch `~/.config/nats`.

## Gates

- `bunx tsc --noEmit` → exit 0
- `bun test` → **720 pass / 2 pre-existing env-dependent fails** (identical set as baseline `main` — both die at `Error: set an operator -- 'nsc env -o operatorName'`; require a configured nsc operator on the local machine). +5 net pass.

## Why this matters

This is the load-bearing piece for any credential lifecycle story across the metafactory ecosystem (cortex sub-bots, future arc-installable bots, any per-component NATS user). cortex#77 was filed because cortex couldn't trust that `arc nats remove-bot` did a real revoke — and now we know why: it didn't. After this PR, the cortex#77 close ("revoke is arc's domain — use `arc nats remove-bot`") is correct end-to-end.

Refs cortex#77 (the consumer that surfaced this).

### DIFF
--- a/src/commands/nats.ts
+++ b/src/commands/nats.ts
@@ -21,10 +21,42 @@ const NSC_CONFIG_CANDIDATES = [
   join(process.env.XDG_CONFIG_HOME ?? join(homedir(), ".config"), "nsc", "nsc.json"),
 ];
 
-function nsc(args: string[]): string {
+/**
+ * Result of an nsc invocation. Shape matches the subset of Bun.spawnSync
+ * we care about; broken out as an interface so tests can stub the runner
+ * without depending on Bun internals.
+ */
+export interface NscResult {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}
+
+export type NscRunner = (args: string[]) => NscResult;
+
+const defaultRunner: NscRunner = (args) => {
   const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
-  const stdout = result.stdout.toString().trim();
-  const stderr = result.stderr.toString().trim();
+  return {
+    exitCode: result.exitCode ?? -1,
+    stdout: result.stdout.toString(),
+    stderr: result.stderr.toString(),
+  };
+};
+
+let runner: NscRunner = defaultRunner;
+
+/**
+ * Test-only hook to swap the nsc runner. Pass `null` to restore the default.
+ * Do NOT call from production code paths.
+ */
+export function __setNscRunnerForTests(next: NscRunner | null): void {
+  runner = next ?? defaultRunner;
+}
+
+function nsc(args: string[]): string {
+  const result = runner(args);
+  const stdout = result.stdout.trim();
+  const stderr = result.stderr.trim();
   if (result.exitCode !== 0) {
     throw new Error(`nsc ${args[0]} failed: ${stderr || stdout || "unknown error"}`);
   }
@@ -32,18 +64,28 @@ function nsc(args: string[]): string {
 }
 
 function nscWithStderr(args: string[]): string {
-  const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
+  const result = runner(args);
   if (result.exitCode !== 0) {
-    const stderr = result.stderr.toString().trim();
-    const stdout = result.stdout.toString().trim();
+    const stderr = result.stderr.trim();
+    const stdout = result.stdout.trim();
     throw new Error(`nsc ${args[0]} failed: ${stderr || stdout || "unknown error"}`);
   }
-  return (result.stdout.toString() + result.stderr.toString()).trim();
+  return (result.stdout + result.stderr).trim();
+}
+
+let nscInstallCheck: () => boolean = () => {
+  return Bun.spawnSync(["which", "nsc"], { stdout: "pipe" }).exitCode === 0;
+};
+
+/**
+ * Test-only hook to swap the install check. Pass `null` to restore the default.
+ */
+export function __setNscInstallCheckForTests(next: (() => boolean) | null): void {
+  nscInstallCheck = next ?? (() => Bun.spawnSync(["which", "nsc"], { stdout: "pipe" }).exitCode === 0);
 }
 
 export function ensureNscInstalled(): void {
-  const result = Bun.spawnSync(["which", "nsc"], { stdout: "pipe" });
-  if (result.exitCode !== 0) {
+  if (!nscInstallCheck()) {
     console.error("Error: nsc not found on PATH.");
     console.error("Install: brew install nats-io/nats-tools/nsc");
     process.exit(1);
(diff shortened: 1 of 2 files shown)


src/commands/nats.ts
```diff
@@ -21,29 +21,71 @@ const NSC_CONFIG_CANDIDATES = [
...
+
+let nscInstallCheck: () => boolean = () => {
+  return Bun.spawnSync(["which", "nsc"], { stdout: "pipe" }).exitCode === 0;
+};
+
+/**
+ * Test-only hook to swap the install check. Pass `null` to restore the default.
+ */
+export function __setNscInstallCheckForTests(next: (() => boolean) | null): void {
+  nscInstallCheck = next ?? (() => Bun.spawnSync(["which", "nsc"], { stdout: "pipe" }).exitCode === 0);
```
mellanon, line 84: **[suggestion/quality]** Symmetry nit: `defaultRunner` is named and reused as the restore default in `__setNscRunnerForTests` (line 53), but `nscInstallCheck`'s default is inlined twice — once at line 76-78 (initial) and once at line 84 (restore). Consider extracting:

```ts
const defaultInstallCheck: () => boolean = () => Bun.spawnSync(["which", "nsc"], { stdout: "pipe" }).exitCode === 0;
let nscInstallCheck: () => boolean = defaultInstallCheck;
export function __setNscInstallCheckForTests(next: (() => boolean) | null): void {
  nscInstallCheck = next ?? defaultInstallCheck;
}
```

Drops the duplicated closure and matches the runner pattern exactly. Non-blocking — pure DRY.

src/commands/nats.ts
```diff
@@ -95,6 +137,61 @@ function userExists(account: string, name: string): boolean {
...
+ * Aborts (throws) on push failure — the caller MUST NOT proceed to delete
+ * the user locally if this throws, because a half-done revoke leaves the
+ * JWT valid on the bus and the operator unaware.
+ */
+function revokeAndPushUser(account: string, name: string): void {
+  const pubKey = getUserPubKey(account, name);
+
+  // Add to revocation map (keyed by pubkey so it survives local user delete).
+  try {
+    nsc(["revocations", "add-user", "-a", account, "-u", pubKey]);
```
mellanon, line 174: **[note/security]** Confirmed: revocation uses `-u <pubkey>` (pubkey-keyed via `getUserPubKey` line 145-158), NOT `-n <name>`. This is the right call — the revocation entry survives the subsequent `nsc delete user` because it's keyed by durable NKey identity, not the local name. Verified by reading the call site + the test at lines 126-129 which explicitly asserts `revokeCall.args` contains `-u` and the captured pubkey.

Pubkey resolution must happen BEFORE delete (`getUserPubKey` is called from `revokeAndPushUser` which runs at line 403 in `removeBot` and line 304 in `reissueBot`, both BEFORE `nsc delete user`). After delete the user record is gone and `nsc describe` would fail. Order is correct.

src/commands/nats.ts
```diff
@@ -95,6 +137,61 @@ function userExists(account: string, name: string): boolean {
...
+  // If this fails, the JWT remains valid server-side — abort loudly.
+  try {
+    nsc(["push", "-a", account]);
+  } catch (err) {
+    const reason = err instanceof Error ? err.message : String(err);
+    throw new Error(
+      `Server-side revoke failed: ${reason}. The user JWT is STILL VALID on the bus. ` +
+      `Resolve connectivity and retry.`
+    );
+  }
```
mellanon, line 192: **[note/security]** Confirmed: abort-on-failure semantics are real. If `push` throws (line 185-192), `revokeAndPushUser` propagates the error up to the caller. Both `removeBot` (lines 402-407) and `reissueBot` (lines 303-308) catch, `console.error` the message (which contains `The user JWT is STILL VALID on the bus`), and `process.exit(1)`. The subsequent `nsc delete user` at line 409 / 310 is never reached.

Test `removeBot — server-side revocation > aborts and does NOT call delete user when push fails` (lines 136-170) proves this: `expect(deleteCalled).toBe(false)` AND `expect(errMessages).toContain("STILL VALID")`. Half-done revoke is impossible on this path.

mellanon (commented):
Full review (6 lenses + duplication) — server-side NATS revoke + push before delete; all 9 reviewer priorities verified by reading the code; gates green (tsc 0, 7/7 new tests, 720/722 with 2 pre-existing env-dependent fails identical to main).

verdict: blockers=0 majors=0 nits=1 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

PR #133, opened 2026-05-12
fix(nats): push server-side revocation in remove-bot/reissue-bot (closes #130)

jcfischer:
## Summary

Fixes #130 — security gap: `arc nats remove-bot` and `arc nats reissue-bot` ran only `nsc delete user`, leaving the user's JWT valid on the bus until natural expiry. Any leaked `.creds` (backup, container image, prior filesystem access, the `.bak` file `reissue` itself wrote) continued to authenticate.

## What changed

Both commands now run the full revoke flow **before** the local delete:

1. `nsc describe user --field sub` — resolve pubkey (must precede delete)
2. `nsc revocations add-user` — add pubkey to account revocations map
3. `nsc push -a <account>` — push updated account JWT to `$SYS`
4. `nsc delete user` — local cleanup

If the server push fails, the operation **aborts** and the user is NOT deleted locally — leaves a recoverable state and surfaces:

```
Server-side revoke failed: <reason>. The user JWT is STILL VALID on the bus. Resolve and retry.
```

## Test seam

Module-private `nsc()` runner is now backed by a swappable `NscRunner` so unit tests can verify call ordering and abort semantics without nsc or a running NATS server. Exposed as `__setNscRunner(fn | null)`.

## Tests

- `test/unit/nats-revocation.test.ts` (8 tests, unconditional):
  - ordering: `describe → revocations add-user → push → delete user` for both `removeBot` and `reissueBot`
  - pubkey from `describe --field sub` is passed to `revocations add-user`
  - quote-stripping on `--field sub` output
  - push failure → throws "Server-side revoke failed … STILL VALID" + does NOT call `delete user`
  - pubkey-lookup failure → aborts with clear error + does NOT call `delete user`
  - revoke happens even when `--delete-creds` is false (independent of local file cleanup)
  - reissue: old creds file remains on disk if push aborts
- `test/commands/nats.test.ts` lifecycle test now gated on a non-destructive `nsc push --diff` probe — skipped when operator JWT has no resolver-server URL.

## Drive-by

`test/commands/identity.test.ts` was leaving `process.exitCode = 1` after two assertion tests because the "restore" line wrote `undefined` back, and `bun:test` inherits the last assigned exit code. Coerced to `0` with `origExitCode ?? 0`. Pre-existing on main; was blocking the pre-push smoke gate.

## Out of scope (follow-ups)

- Real integration test against a running `nats-server` with resolver — the issue's "actually rejects subsequent connections" check needs a CI fixture (`nats-server` container with operator JWT pre-configured). File as separate issue.
- `--json` output mode for cortex delegation (issue #131).

## Deferred edge case

The issue raises the question of whether `nsc push` needs a specific operator/account context. Today's flow inherits the current operator from `nsc env`, same as `addBot`. If a deployment uses a non-default operator the operator should `nsc env -o <op>` first — same precondition as the rest of the `arc nats` surface.

## Version

`0.24.0` → `0.24.1` (patch — bug fix, no API surface change beyond adding `__setNscRunner` for test injection).

## Test plan

- [x] `bun test` — 722 pass, 1 skip (lifecycle test on hosts without resolver-enabled NATS), 0 fail
- [x] `bun test test/unit/nats-revocation.test.ts` — 8 pass
- [ ] Manual: `arc nats add-bot foo` → `arc nats remove-bot foo` against a resolver-enabled local nats-server, verify `nats --creds=./foo.creds.bak pub …` is rejected post-revoke

Closes #130

### DIFF
--- a/src/commands/nats.ts
+++ b/src/commands/nats.ts
@@ -21,10 +21,34 @@ const NSC_CONFIG_CANDIDATES = [
   join(process.env.XDG_CONFIG_HOME ?? join(homedir(), ".config"), "nsc", "nsc.json"),
 ];
 
-function nsc(args: string[]): string {
+export interface NscResult {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}
+
+export type NscRunner = (args: string[]) => NscResult;
+
+const realNscRunner: NscRunner = (args) => {
   const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
-  const stdout = result.stdout.toString().trim();
-  const stderr = result.stderr.toString().trim();
+  return {
+    exitCode: result.exitCode ?? 1,
+    stdout: result.stdout.toString(),
+    stderr: result.stderr.toString(),
+  };
+};
+
+let currentNscRunner: NscRunner = realNscRunner;
+
+// Test seam: replace the nsc invoker. Pass null to restore the real runner.
+export function __setNscRunner(fn: NscRunner | null): void {
+  currentNscRunner = fn ?? realNscRunner;
+}
+
+function nsc(args: string[]): string {
+  const result = currentNscRunner(args);
+  const stdout = result.stdout.trim();
+  const stderr = result.stderr.trim();
   if (result.exitCode !== 0) {
     throw new Error(`nsc ${args[0]} failed: ${stderr || stdout || "unknown error"}`);
   }
@@ -32,13 +56,13 @@ function nsc(args: string[]): string {
 }
 
 function nscWithStderr(args: string[]): string {
-  const result = Bun.spawnSync(["nsc", ...args], { stderr: "pipe", stdout: "pipe" });
+  const result = currentNscRunner(args);
   if (result.exitCode !== 0) {
-    const stderr = result.stderr.toString().trim();
-    const stdout = result.stdout.toString().trim();
+    const stderr = result.stderr.trim();
+    const stdout = result.stdout.trim();
     throw new Error(`nsc ${args[0]} failed: ${stderr || stdout || "unknown error"}`);
   }
-  return (result.stdout.toString() + result.stderr.toString()).trim();
+  return (result.stdout + result.stderr).trim();
 }
 
 export function ensureNscInstalled(): void {
@@ -92,6 +116,46 @@ function userExists(account: string, name: string): boolean {
     return true;
   } catch {
     return false;
+  }
+}
+
+/**
+ * Server-side revocation: add the user's public key to the account's revocations
+ * map, then push the updated account JWT to the NATS server's $SYS account so
+ * the server starts rejecting connections from any creds derived from that key.
+ *
+ * MUST be called before `nsc delete user` — the pubkey can't be looked up after
+ * the local user record is gone. Throws on push failure WITHOUT deleting the
+ * user, leaving a recoverable state.
+ */
+function revokeAndPush(account: string, name: string): void {
+  // 1. Resolve pubkey from the user JWT (`sub` field). Must happen before delete.
+  let userPubKey: string;
+  try {
+    userPubKey = nsc(["describe", "user", "-a", account, "-n", name, "--field", "sub"]);
+  } catch (err) {
+    throw new Error(
+      `Cannot look up pubkey for user "${name}" — server-side revoke skipped. ` +
+      `Cause: ${err instanceof Error ? err.message : String(err)}`,
+    );
+  }
+  // nsc may wrap `--field` output in quotes; strip them.
+  userPubKey = userPubKey.replace(/^"|"$/g, "");
+  if (!userPubKey) {
+    throw new Error(`Empty pubkey returned for user "${name}" — refusing to proceed.`);
+  }
+
+  // 2. Add to account revocations map.
+  nsc(["revocations", "add-user", "-a", account, "-u", userPubKey]);
+
+  // 3. Push the updated account JWT to the NATS server's $SYS account.
+  try {
+    nsc(["push", "-a", account]);
+  } catch (err) {
+    throw new Error(
+      `Server-side revoke failed: ${err instanceof Error ? err.message : String(err)}. ` +
+      `The user JWT is STILL VALID on the bus. Resolve and retry.`,
+    );
   }
 }
 
@@ -201,6 +265,12 @@ export function reissueBot(name: string, opts: ReissueBotOptions): void {
     console.log(`  backup: ${backup}`);
   }
 
+  // Server-side revoke of the OLD user's pubkey BEFORE delete. The new user
+  // gets a fresh keypair and is unaffected. If push fails, abort without
+  // delete (recoverable). See issue #130.
+  revokeAndPush(account, name);
+  console.log(`Revoked old credentials on server`);
+
   nsc(["delete", "user", "-a", account, "-n", name]);
 
   try {
@@ -211,7 +281,7 @@ export function reissueBot(name: string, opts: ReissueBotOptions): void {
     console.error(`CRITICAL: failed to re-create user "${name}" after delete.`);
     if (existsSync(`${outPath}.bak`)) {
       console.error(`Old credentials backed up at: ${outPath}.bak`);
-      console.error(`Note: old creds are invalidated — backup is for reference only.`);
+      console.error(`Note: old creds are already revoked server-side — backup is for forensic reference only.`);
     }
     console.error(`Manual recovery: nsc add user -a ${account} -n ${name}`);
     console.error(`Cause: ${err instanceof Error ? err.message : String(err)}`);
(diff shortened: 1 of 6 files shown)


src/commands/nats.ts
```diff
@@ -201,6 +265,12 @@ export function reissueBot(name: string, opts: ReissueBotOptions): void {
     console.log(`  backup: ${backup}`);
   }
```
mellanon, line 266: **[major/security]** `.bak` is written BEFORE `revokeAndPush` (line 263 vs line 271). If the push aborts, the `.bak` file stays on disk holding the **still-valid** old creds — and `reissueBot`'s own test (`old creds file remains on disk if push aborts`) confirms this is the post-PR behaviour.

This re-creates the exact threat model from #130: a leaked backup that authenticates against the bus. Issue #130 explicitly notes *"The .bak file is currently written BEFORE the revocation push. After this fix, the .bak file is automatically dead too"* — that invariant only holds on the success path here. On push failure, the `.bak` exists and is still live.

**Fix options (pick one):**
1. Move the `.bak` write to AFTER `revokeAndPush` succeeds (preferred — preserves the "backup only exists when the key is dead" invariant).
2. Keep the current order, but delete the `.bak` in the push-failure path before re-throwing.

Fail-closed means the failure path must not leave live credentials on disk.

src/commands/nats.ts
```diff
@@ -21,24 +21,48 @@ const NSC_CONFIG_CANDIDATES = [
...
+    exitCode: result.exitCode ?? 1,
+    stdout: result.stdout.toString(),
+    stderr: result.stderr.toString(),
+  };
+};
+
+let currentNscRunner: NscRunner = realNscRunner;
+
+// Test seam: replace the nsc invoker. Pass null to restore the real runner.
+export function __setNscRunner(fn: NscRunner | null): void {
```
mellanon, line 44: **[suggestion/hardening]** `__setNscRunner` is a module-level `export function`, so any caller that imports `arc/dist/commands/nats.js` can globally swap the runner process-wide. The `__` prefix is convention only — not enforced.

Not a vulnerability today (the CLI doesn't re-export it, and arc isn't consumed as a library by untrusted code), but the test seam is mutable global state. Two defensive options:

1. Gate it on env: early-return unless `process.env.NODE_ENV === "test"` or `process.env.ARC_TEST_MODE === "1"`. Cheap, makes a production no-op runner injection impossible.
2. Move the runner to a thin object exported from a `__test__` subpath (or a separate `nats.testing.ts`) so it doesn't sit on the public surface of the production module.

Option 1 is one line and gets the job done.

src/commands/nats.ts
```diff
@@ -21,24 +21,48 @@ const NSC_CONFIG_CANDIDATES = [
...
-  const stdout = result.stdout.toString().trim();
-  const stderr = result.stderr.toString().trim();
+  return {
+    exitCode: result.exitCode ?? 1,
+    stdout: result.stdout.toString(),
+    stderr: result.stderr.toString(),
+  };
+};
+
+let currentNscRunner: NscRunner = realNscRunner;
```
mellanon, line 41: **[suggestion/architecture]** `currentNscRunner` is process-wide mutable state. The tests serialise correctly via `afterEach(() => __setNscRunner(null))`, but if `bun test` is ever run with parallelism inside the same process (or two suites are imported into one test file), runner swaps will race.

Low probability today; flagging because the seam is the kind of thing that grows tests over time. If it ever becomes a problem, the fix is to thread an optional `runner` arg through `revokeAndPush(account, name, runner?)` and have the tests pass it explicitly — no module-level state.

src/commands/nats.ts
```diff
@@ -95,6 +119,46 @@ function userExists(account: string, name: string): boolean {
...
+
+  // 2. Add to account revocations map.
+  nsc(["revocations", "add-user", "-a", account, "-u", userPubKey]);
+
+  // 3. Push the updated account JWT to the NATS server's $SYS account.
+  try {
+    nsc(["push", "-a", account]);
+  } catch (err) {
+    throw new Error(
+      `Server-side revoke failed: ${err instanceof Error ? err.message : String(err)}. ` +
```
mellanon, line 156: **[nit/security]** The wrapped error string inlines the raw `nsc` stderr verbatim. `nsc push` failure messages can include the resolver URL, server hostname, and connection-level details. For an operator-run CLI that's fine; just be aware this lands in any log aggregator that ingests stderr from these commands.

No action required if logs are operator-scoped. If they ever go further (e.g. piped to a shared incident channel), redact the cause.

src/commands/nats.ts
```diff
@@ -201,6 +265,12 @@ export function reissueBot(name: string, opts: ReissueBotOptions): void {
     console.log(`  backup: ${backup}`);
   }
 
+  // Server-side revoke of the OLD user's pubkey BEFORE delete. The new user
+  // gets a fresh keypair and is unaffected. If push fails, abort without
+  // delete (recoverable). See issue #130.
+  revokeAndPush(account, name);
+  console.log(`Revoked old credentials on server`);
```
mellanon, line 272: **[nit/quality]** Inconsistent log shape: `removeBot` logs `Revoked on server: ${name} (account: ${account})` (line 361), this logs just `Revoked old credentials on server`. Match the format so log scrapers / audit greps catch both call sites with one pattern.

```ts
console.log(`Revoked on server: ${name} (account: ${account})`);
```

test/unit/nats-revocation.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        };
+      })(),
+    });
+    __setNscRunner(run);
+
+    expect(() => removeBot(BOT, { account: ACCOUNT })).toThrow(/server-side revoke skipped/);
+
+    const deleteCall = calls.find((c) => c.args[0] === "delete" && c.args[1] === "user");
+    expect(deleteCall).toBeUndefined();
+  });
```
mellanon, line 143: **[nit/quality]** The "pubkey-lookup failure" test relies on internal sequencing: it fails the *second* describe call (the one with `--field`) and assumes `userExists()` issues the *first* describe. If `userExists` is ever rewritten to use a different probe (e.g. `nsc list users | grep`), this test passes vacuously without exercising the abort-on-lookup-fail path.

Tighter: match on `args.includes("--field")` directly without a seen-counter — the `--field` flag is the unique marker of the lookup call.

```ts
fail: (args) => (args[0] === "describe" && args.includes("--field") ? "user JWT not found" : null),
```

test/commands/identity.test.ts
```diff
@@ -159,7 +163,11 @@ describe("importPrincipals — security", () => {
     const origExitCode = process.exitCode;
     importPrincipals(importFile);
     expect(process.exitCode).toBe(1);
-    process.exitCode = origExitCode;
+    // Setting `process.exitCode = undefined` does NOT clear it under Bun —
+    // the process retains the last assigned value, which makes the whole
+    // `bun test` run exit non-zero even with 0 fails. Coerce to 0 so the
+    // smoke-test gate sees a clean exit.
+    process.exitCode = origExitCode ?? 0;
```
mellanon, line 170: **[nit/duplication]** Same 4-line `process.exitCode = origExitCode ?? 0` block with identical comment appears at lines 150-153 and 166-169 of this file. One helper would do it:

```ts
function resetExitCode(orig: number | undefined) {
  process.exitCode = orig ?? 0;
}
```

Minor; drive-by scope is fine to leave as-is, but flagging for future cleanup.

mellanon (changes requested):
Full review (6 lenses + duplication) on #133 — server-side NATS revocation.

The core fix is correct: `revokeAndPush` runs BEFORE `nsc delete user` in both flows, push failure throws with a clear "STILL VALID" message, and the local delete is skipped on abort. Pubkey lookup ordering, quote-stripping, and the `--delete-creds`-independent revoke path are all tested. Good test seam pattern; mock-based unit tests don't require nsc or a NATS server.

**One blocker for the issue's threat model:** in `reissueBot`, the `.bak` file is written BEFORE `revokeAndPush`. On push failure the operation aborts (correct) — but leaves the `.bak` on disk holding the still-valid old creds. That's exactly the leaked-backup vector #130 was filed to close. Fix is one of: write `.bak` after `revokeAndPush` succeeds (preferred), or delete `.bak` in the push-failure path. The PR's own test asserts the current (unsafe) behaviour, so flipping it is a test change too.

**Lens results**

| Lens | Critical | Major | Suggestion | Nit | Verdict |
|------|----------|-------|------------|-----|---------|
| CodeQuality | 0 | 0 | 0 | 2 | pass |
| Security | 0 | 1 | 0 | 1 | fail |
| Hardening | 0 | 0 | 1 | 0 | advisory |
| Architecture | 0 | 0 | 1 | 0 | pass |
| EcosystemCompliance | 0 | 0 | 0 | 0 | pass |
| Performance | 0 | 0 | 0 | 0 | pass |
| Duplication | 0 | 0 | 0 | 1 | pass |

**Other findings (non-blocking)**
- `__setNscRunner` is a module-level export — gate on `NODE_ENV === "test"` to prevent any production runner swap.
- `currentNscRunner` is process-wide mutable state; tomorrow's parallel test runners will race.
- Inconsistent log format between `removeBot` and `reissueBot` revoke lines (audit-grep impact).
- The pubkey-lookup-failure test depends on `userExists` issuing exactly one prior `describe` — match on `--field` flag instead of a seen-counter.
- Drive-by `process.exitCode ?? 0` block duplicated; one helper would tidy it.

Version bump (0.24.0 → 0.24.1, patch) and commit prefix (`fix(nats):`) are correct. PR carries no labels — add `bug` + `now` per repo standard.

verdict: blockers=0, majors=1, nits=5 — recommend: request-changes

PR #139, opened 2026-05-13
feat(remove): cleanup parity with install — provides.files + preremove hook + -y flag (refs arc#138)

mellanon:
## Summary

Brings `arc remove` closer to `arc install` parity. Discovered during the cortex v1 operator-mode cutover (arc#138): removing Grove left launchd plists, dangling `~/bin` + `~/.claude/hooks/` symlinks, settings.json registrations, and running daemons behind.

This PR closes the smallest, highest-leverage subset of arc#138:

- **A. provides.files mirror cleanup.** `arc remove` now reverse-iterates `manifest.provides.files[]` and unlinks each target IFF it is still a symlink AND `readlink(target)` matches the source we installed. Hand-edited files (regular files, or symlinks pointing elsewhere) are left untouched with a `console.warn`. arc never deletes operator content.

- **B. `scripts.preremove` lifecycle hook.** New slot in `ArcManifest.scripts`. Fires **before** any teardown so the package can stop its own daemons, unload launchd plists, etc. while its symlinks still resolve. Missing script falls through via `runScript`'s `skipped` path (same shape as preinstall/preupgrade). A failing preremove warns but does **not** abort — operators need cleanup to finish even when a daemon is already dead.

- **D1. `-y, --yes` flag on `arc remove`.** Mirrors `arc install -y`. Plumbed through `remove(...)` and `removeLibrary(...)` as a `RemoveOptions` object so existing call sites stay back-compat.

## Closes from arc#138

- [x] **A. Mirror `provides.files[]` on remove** (with the `readlink`-matches-source safety check)
- [x] **B. Add `preremove` hook** (`postremove` not added — install/upgrade share `postinstall` fallback semantics; remove has no analogous case where we'd want post-cleanup, since the repo is already gone)
- [x] **D1. `arc remove -y`**

## Deferred to follow-ups

- **C. Sweep `~/.claude/settings.json`** — Going-forward path (every install tags `_pai_pkg`) is already in place. The hard part is **untagged legacy entries** (Grove had 42 of these). Needs design conversation — auto-deleting hook entries matching a `/<Name>` prefix is risky for operators who hand-curated their settings. Out of scope for this PR per arc#138's "biggest PR, don't bloat" guidance.

- **D2. `arc install` postinstall investigation** — Cortex's manifest declares `scripts.postinstall: ./scripts/postinstall.sh` correctly (verified — the key is `scripts:` not `hooks:`, and the script exists in the repo). The install pipeline at `src/commands/install.ts:351` calls `runScript` with the correct path. The `scripts.postinstall` feature shipped in arc v0.10.0, well before the v1 cutover.

  **Best guess from the source trace:** with `-y`, `runScript` is called with `quiet: true`, which pipes the script's stdout/stderr instead of inheriting them (`scripts.ts:48-50`). If `cortex/scripts/postinstall.sh` errored silently (e.g. `set -e` fired before any visible output, or a stale `PAI_INSTALL_PATH` resolved wrong), the operator would see no log and no plists rendered, indistinguishable from "the script never ran." Closing this gap probably means surfacing the postinstall exit code in `-y` mode regardless. Filing as a separate UX issue.

- **D3. `github:org/repo` URL shorthand** — separate UX issue, no source-resolver entry currently handles the `github:` prefix.

## Test plan

- [x] `bun test test/commands/remove.test.ts` — 14 pass (8 existing + 6 new)
- [x] `bun test` — 746 pass / 3 skip / 0 fail across 47 files
- [x] `bunx tsc --noEmit` — clean

New tests in `test/commands/remove.test.ts > "remove parity with install (arc#138)"`:
1. `-y` flag accepted
2. `provides.files` targets removed when symlink+target match
3. Hand-edited files (regular file replacing arc-installed symlink) left untouched
4. Redirected symlinks (operator re-pointed elsewhere) left untouched
5. `scripts.preremove` fires BEFORE teardown — verified by the script checking the symlink still exists when it runs
6. Absent `scripts.preremove` → silent no-op

## Drive-by

`test/commands/nats.test.ts` had two CLI-subject-validation tests that crashed when the active `nsc` operator didn't carry the hardcoded `OP_JC` account (`OP_ANDREAS` on this machine doesn't). Tightened the skip guard to also probe `nsc list accounts` for `TEST_ACCOUNT`. Pre-existing flake unrelated to arc#138, but blocked the full-suite green-bar.

### DIFF
--- a/src/commands/remove.ts
+++ b/src/commands/remove.ts
@@ -1,5 +1,6 @@
 import { join } from "path";
-import { rm } from "fs/promises";
+import { rm, lstat, readlink, unlink } from "fs/promises";
+import { homedir } from "os";
 import type { Database } from "bun:sqlite";
 import type { ArcPaths, HostAdapter } from "../types.js";
 import { getSkill, removeSkill, listByLibrary } from "../lib/db.js";
@@ -8,6 +9,12 @@ import { readManifest } from "../lib/manifest.js";
 import { removeHooks, hasHooks } from "../lib/hooks.js";
 import { findGitRoot } from "../lib/paths.js";
 import { unwireExtensions } from "../lib/extensions.js";
+import { runScript } from "../lib/scripts.js";
+
+export interface RemoveOptions {
+  /** Suppress interactive prompts / informational output (for non-interactive / test use). */
+  yes?: boolean;
+}
 
 export interface RemoveResult {
   success: boolean;
@@ -18,17 +25,55 @@ export interface RemoveResult {
 
 /**
  * Completely remove an installed skill.
- * Removes symlinks, repo directory, and database entry.
+ *
+ * Mirrors `install` in reverse:
+ *  1. Fire `scripts.preremove` (if declared) so the package can stop daemons
+ *     and clean up host-side state (launchd plists, etc.) BEFORE its repo +
+ *     symlinks are torn down. See arc#138.
+ *  2. Tear down primary symlink (skill/agent/tool/...) and CLI shims.
+ *  3. Remove hooks from settings.json and unwire extensions.
+ *  4. Reverse-iterate `provides.files` and unlink each target IFF it is a
+ *     symlink pointing at the source we installed (hand-edited files are
+ *     left in place with a warning).
+ *  5. Delete the DB row and the cloned repo directory.
  */
 export async function remove(
   db: Database,
   arc: ArcPaths,
   host: HostAdapter,
-  name: string
+  name: string,
+  opts: RemoveOptions = {}
 ): Promise<RemoveResult> {
   const skill = getSkill(db, name);
   if (!skill) {
     return { success: false, error: `Skill '${name}' is not installed` };
+  }
+
+  // Read manifest up-front. Needed for preremove firing, provides.files
+  // cleanup, CLI shim names, and hooks teardown. Best-effort: if the manifest
+  // is unreadable (corrupted clone, manual rm -rf, etc.) we fall through with
+  // a null manifest and skip the manifest-driven cleanup steps — the DB row
+  // and primary symlink will still be removed.
+  const manifest = await readManifest(skill.install_path).catch(() => null);
+
+  // 1. Fire preremove script BEFORE anything is torn down. Same shape as
+  //    preinstall/preupgrade — gives the package a chance to stop daemons,
+  //    unload launchd plists, etc. while its files and symlinks still exist.
+  //    Missing script falls through silently via runScript's skipped path.
+  if (manifest?.scripts?.preremove) {
+    const preResult = runScript({
+      installPath: skill.install_path,
+      scriptPath: manifest.scripts.preremove,
+      hookName: "preremove",
+      quiet: opts.yes,
+    });
+    if (!preResult.success && !preResult.skipped) {
+      // Do NOT abort: a failing preremove (e.g. daemon already stopped) must
+      // not block the user from finishing the cleanup. Warn loudly instead.
+      console.warn(
+        `  ⚠ preremove script exited ${preResult.exitCode}; continuing remove anyway`,
+      );
+    }
   }
 
   const isTool = skill.artifact_type === "tool";
@@ -69,9 +114,6 @@ export async function remove(
     await removeSymlink(skillLink);
   }
 
-  // Read manifest before removal (needed for CLI shim names and hooks cleanup)
-  const manifest = await readManifest(skill.install_path);
-
   // Remove all CLI shims and bin symlinks (skills and tools)
   if (!isAgent && !isPrompt && manifest) {
     const cliEntries = extractAllCliInfo(manifest);
@@ -96,6 +138,23 @@ export async function remove(
   // Remove extensions (before deleting repo)
   if (manifest?.extensions) {
     await unwireExtensions(manifest, host.paths.root);
+  }
+
+  // Mirror provides.files on the way out.
+  //
+  // For every {source, target} the installer created, unlink the target IFF
+  // it is still a symlink AND it points at the source path we installed.
+  //
+  // Safety rule: never `rm` a target that has been hand-edited (i.e. it's
+  // a regular file now, or its symlink target no longer matches). The
+  // operator may have replaced an arc-installed file with their own copy;
+  // arc has no business deleting that. Warn instead. See arc#138.
+  if (manifest?.provides?.files?.length) {
+    for (const file of manifest.provides.files) {
+      const expectedSource = join(skill.install_path, file.source);
+      const targetPath = file.target.replace(/^~/, homedir());
+      await removeProvidedFile(targetPath, expectedSource);
+    }
   }
 
   // Remove from database (CASCADE deletes capabilities) — before repo removal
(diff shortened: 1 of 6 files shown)


src/commands/remove.ts
```diff
@@ -122,6 +181,58 @@ export async function remove(
...
+
+  let actualTarget: string;
+  try {
+    actualTarget = await readlink(targetPath);
+  } catch (err: any) {
+    console.warn(`  ⚠ provides.files cleanup: readlink failed on ${targetPath}: ${err?.message ?? err}`);
+    return;
+  }
+
+  if (actualTarget !== expectedSource) {
```
mellanon, line 220: [suggestion/security] The `actualTarget !== expectedSource` byte-equality check is correctly conservative: both install (`src/lib/artifact-installer.ts:275-278`) and remove compute `join(installDir, file.source)` the same way, and `fs.symlink()` writes the target byte-for-byte (Bun + libuv), so `readlink()` returns the same absolute string we passed in. Verified `path.join` normalizes `/abs/install` vs `/abs/install/` vs `/abs/install` + `./files/x` identically. Trailing slashes / relative-vs-absolute drift between install and remove are not a concern here.

The one ambient hazard (out of scope for this PR but worth filing): `path.join(installDir, file.source)` does not guard against a malicious manifest with `source: "../../../etc/passwd"`. `path.join` collapses traversal: `join("/Users/x/.config/metafactory/pkg/repos/Mal", "../../../etc/passwd")` -> `/Users/x/.config/metafactory/etc/passwd`. Install would create the symlink to that path; remove would faithfully unlink it. Mitigation belongs in install (assert resolved sourcePath stays under installDir). Suggest filing as follow-up.

src/commands/remove.ts
```diff
@@ -98,6 +140,23 @@ export async function remove(
...
+  // a regular file now, or its symlink target no longer matches). The
+  // operator may have replaced an arc-installed file with their own copy;
+  // arc has no business deleting that. Warn instead. See arc#138.
+  if (manifest?.provides?.files?.length) {
+    for (const file of manifest.provides.files) {
+      const expectedSource = join(skill.install_path, file.source);
+      const targetPath = file.target.replace(/^~/, homedir());
+      await removeProvidedFile(targetPath, expectedSource);
+    }
+  }
```
mellanon, line 158: [suggestion/ux] Atomicity: each `removeProvidedFile` swallows its own errors via `console.warn` and continues, which is the right call (operator must always be able to finish a remove). But the `RemoveResult` does not surface a partial-cleanup count to the CLI. Consider extending `RemoveResult` with `leftBehind?: string[]` populated when `removeProvidedFile` skips a target, so `arc remove -y` can finish with a non-fatal summary like `Removed Foo (2 files left in place — see warnings above)`. Without that, an operator using `-y` (quiet mode) may not notice that hand-edited customisations exist that they should review. Not a blocker; flag as next-iteration.

src/commands/remove.ts
```diff
@@ -18,19 +25,57 @@ export interface RemoveResult {
...
+  //    preinstall/preupgrade — gives the package a chance to stop daemons,
+  //    unload launchd plists, etc. while its files and symlinks still exist.
+  //    Missing script falls through silently via runScript's skipped path.
+  if (manifest?.scripts?.preremove) {
+    const preResult = runScript({
+      installPath: skill.install_path,
+      scriptPath: manifest.scripts.preremove,
+      hookName: "preremove",
+      quiet: opts.yes,
+    });
```
mellanon, line 69: [note/architecture] Verified `preremove` shares the `runScript` infra with `preinstall`/`postinstall`/`preupgrade`. Env vars passed to the script are `PAI_INSTALL_PATH=<installPath>` and `PAI_HOOK="preremove"` (`src/lib/scripts.ts:52-56`). This is symmetric with the other lifecycle hooks — `cortex/scripts/preupgrade.sh` and any future `preremove.sh` will see the same env. Good.

One subtle gap: `runScript` calls `Bun.spawnSync` with `cwd: installPath`, so the script runs from the package root with the package files still resolvable. The PR comment "while symlinks still resolve" is verified by `test/commands/remove.test.ts:289-303` (the marker file records `preremove-pre-teardown`). Solid test design.

test/commands/remove.test.ts
```diff
@@ -168,3 +169,160 @@ describe("removeLibrary", () => {
...
+      name: "YesFlag",
+    });
+    await install({
+      arc: env.arc, host: env.host,
+      db: env.db,
+      repoUrl: repo.url,
+      yes: true,
+    });
+
+    // Smoke: opts.yes is accepted without throwing, mirrors install.
```
mellanon, line 185: [suggestion/test-coverage] The 6 new tests cover the documented happy-path + 2 safety edges (regular-file-replacement, redirected-symlink). What is not exercised: the `~/` path expansion in `file.target`. Tests use absolute `env.root`-rooted paths, so the `replace(/^~/, homedir())` branch in both install (`artifact-installer.ts:276`) and remove (`remove.ts:155`) is not directly tested. Practically this is the same code path on both sides so byte-equality holds, but adding one test with `target: "~/fake-home-tilde-test/X"` would close the coverage hole. Not a blocker.

test/commands/nats.test.ts
```diff
@@ -9,6 +9,19 @@ const TEST_ACCOUNT = "OP_JC";
...
+// isn't registered under whoever the active operator happens to be. The
+// error message in that case has nothing to do with subject validation, so
+// these tests are environment-skewed unless the operator carries OP_JC.
+// Skip cleanly when the account isn't available rather than asserting on
+// an unrelated nsc failure (was producing flake — see arc#138 sweep).
+const TEST_ACCOUNT_AVAILABLE = (() => {
+  if (!NSC_AVAILABLE) return false;
+  const probe = Bun.spawnSync(["nsc", "list", "accounts"], { stdout: "pipe", stderr: "pipe" });
+  return probe.exitCode === 0 && probe.stdout.toString().includes(TEST_ACCOUNT);
+})();
```
mellanon, line 23: [note/test-hygiene] Drive-by skip-guard confirmed legitimate. The two affected tests assert on `Invalid NATS subject` in stderr, but if the operator does not carry `OP_JC`, `nsc add user` errors with `account not in operator` BEFORE the subject-validation logic runs — so the assertion fails on an unrelated nsc error. Probing `nsc list accounts | grep OP_JC` correctly skips only the env-skewed cases without hiding real coverage. Pre-existing flake, not a coverage hide.

mellanon (commented):
Echo review — recommend: **merge**.

Symlink validation is correct (byte-equality between install-time and remove-time `join(installDir, file.source)` verified); operator hand-edited content is protected by the conservative skip-on-mismatch path. `preremove` lifecycle hook reuses `runScript` cleanly with the same env vars as other hooks. `-y` plumbing through `RemoveOptions` keeps back-compat. 6 new tests cover happy path + the two most important safety edges. `nats.test.ts` skip-guard is a legitimate env-skew fix, not coverage hiding.

4 non-blocking suggestions posted inline (security follow-up file, atomicity surfacing in CLI, test coverage gap, preremove env validation). PR body’s postinstall investigation finding should be filed as a separate arc issue.

blockers=0 majors=0 suggestions=4

PR #148, opened 2026-05-13
fix(init): in-place scaffold + drop arc-skill- prefix (closes #107)

### LINKED ISSUE #107: arc init: argless rejects, nests inside arc-skill-<name> even in cwd, '.'  becomes literal in dir name
## Problem

`arc init` has three closely-related defects around init-in-place:

### 1. Argless `arc init` rejects with "missing required argument 'name'"

```
~/work/mf/test-blueprint $ arc init
error: missing required argument 'name'
```

Most scaffolders (`npm init`, `cargo init`, `git init`) default the package name to the current directory's basename when run argless. Operators who already `mkdir foo && cd foo` expect `arc init` to scaffold in place.

### 2. `arc init <name>` nests inside an `arc-skill-<name>` subdir even when the cwd already has the right name

```
~/work/mf/test-blueprint $ arc init test-blueprint
✅ Scaffolded skill at ./arc-skill-test-blueprint
Files created:
  arc-manifest.yaml
  skill/SKILL.md
  skill/workflows/Main.md
  package.json
  README.md
  .gitignore
```

I'm already in `test-blueprint/`. The expected outcome is to scaffold those six files at `./` — instead they land in `./arc-skill-test-blueprint/`, which I have to either move up or `cd` into.

The `arc-skill-` prefix and the auto-nesting are also surprising independently:
- The dir name on disk doesn't have to match the manifest's package name. `arc-skill-` adds churn for no benefit a developer can see.
- Even with a fresh `mkdir foo && cd foo && arc init foo`, the result is `foo/arc-skill-foo/` — two `foo`s deep — which is almost never what's wanted.

### 3. `arc init .` produces a literal `arc-skill-.` directory

```
~/work/mf/test-blueprint $ arc init .
✅ Scaffolded skill at ./arc-skill-.
Files created:
  arc-manifest.yaml
  ...

~/work/mf/test-blueprint $ ll
drwxr-xr-x@ 7 fischer  staff   224B Apr 28 07:28 arc-skill-.
```

`.` is the universal "current directory" sigil. `arc init .` should mean "init in this directory". Instead the literal `.` is concatenated into the dir name, producing `arc-skill-.` — the worst of both worlds: not init-in-place, and a dir name that's hard to delete (`rm arc-skill-.` fails because it's a directory; needs `rm -rf` with shell-quoting care).

## Repro

```bash
cd $(mktemp -d)
mkdir test-blueprint
cd test-blueprint

# Bug 1
arc init
# error: missing required argument 'name'

# Bug 2
arc init test-blueprint
# Scaffolded skill at ./arc-skill-test-blueprint  (nested, not in-place)

# Bug 3
arc init .
# Scaffolded skill at ./arc-skill-.  (literal `.` in dir name)
```

## Proposed behavior

- `arc init` with no args, when run in an empty (or matching) cwd, scaffolds in place using the cwd's basename as the package name.
- `arc init <name>` when run in a dir whose basename matches `<name>` scaffolds in place.
- `arc init <name>` when run in a different dir creates `./<name>/` (no `arc-skill-` prefix).
- `arc init .` is equivalent to `arc init` (no args, in place).
- The `arc-skill-` prefix is dropped or moved behind a `--prefix` flag.

## Impact

Hit while setting up a test blueprint to run journey 02 from meta-factory's F-412 journey runner. Three different invocations all produced surprising output before the fourth (`mkdir foo; cd ..; arc init foo`) gave the expected result. Friction for first-time publishers — the same group the whole publishing path is being built for.

## Versions

- arc: (whichever's currently on PATH; output of `arc --version` would help if maintainers want to pin)
- Platform: macOS Darwin 25.3.0

jcfischer:
## Summary

Closes #107 — three closely-related \`arc init\` defects around init-in-place.

| # | Before | After |
|---|--------|-------|
| 1 | \`arc init\` rejects with \`missing required argument 'name'\` | argless → in-place using \`basename(cwd)\` |
| 2 | \`arc init test-blueprint\` from \`test-blueprint/\` nests at \`./arc-skill-test-blueprint/\` | in-place (no nested dir, no \`arc-skill-\` prefix) |
| 3 | \`arc init .\` produces \`./arc-skill-.\` | identical to argless (in-place) |

## Resolution matrix (new pure function \`resolveInitTarget\`)

| argName | cwd | result |
|---------|-----|--------|
| (none) | \`/x/foo\` | name=\`foo\`, targetDir=\`/x/foo\`, inPlace=true |
| \`.\` | \`/x/foo\` | name=\`foo\`, targetDir=\`/x/foo\`, inPlace=true |
| \`foo\` | \`/x/foo\` | name=\`foo\`, targetDir=\`/x/foo\`, inPlace=true |
| \`bar\` | \`/x/foo\` | name=\`bar\`, targetDir=\`/x/foo/bar\`, inPlace=false |
| \`--dir /p\` (any) | (any) | targetDir=\`/p\` |
| path-traversal name | (any) | \`null\` (caller exits with error) |

## init() contract change

- No longer rejects an existing target directory.
- Refuses only when \`arc-manifest.yaml\` already lives at the target.
- \`mkdir(targetDir, { recursive: true })\` runs unconditionally — covers both fresh-dir and in-place cases.

## Tests

- \`test/unit/init-resolve-107.test.ts\` (11) — matrix rows + \`--dir\` precedence + traversal rejection + reproduction scenarios as named cases.
- \`test/commands/init.test.ts\` updated: \"rejects existing directory\" replaced with \"scaffolds into existing non-empty directory\" + \"rejects when arc-manifest.yaml already exists\".

Full suite: **849 / 0** (was 838).

## Smoke verified

\`\`\`
mkdir foo; cd foo; arc init       → ./arc-manifest.yaml, ./skill/...
mkdir foo; cd foo; arc init .     → ./arc-manifest.yaml, ./skill/...
mkdir foo; cd foo; arc init foo   → ./arc-manifest.yaml, ./skill/...
cd /tmp; arc init MyBot           → /tmp/MyBot/arc-manifest.yaml
\`\`\`

## Out of scope

\`package.json\`'s \`name\` field still uses \`arc-<type>-<lowername>\` — that's the published npm package name, distinct from the directory layout the issue complained about.

## Test plan

- [x] \`bun test\` — 849/0
- [x] Three #107 reproduction commands smoke-tested via the worktree binary
- [x] \`arc init\` in a dir with pre-existing files layers cleanly without overwriting

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -609,8 +609,8 @@ program
         cwd: process.cwd(),
         dirOverride: opts.dir,
       });
-      if (!resolved) {
-        console.error(`\n❌ Invalid name. Name must not contain path separators or "..".`);
+      if ("reason" in resolved) {
+        console.error(`\n❌ ${resolved.detail}`);
         process.exit(1);
       }
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -609,7 +609,7 @@ program
         cwd: process.cwd(),
         dirOverride: opts.dir,
       });
-      if ("reason" in resolved) {
+      if (!resolved.ok) {
         console.error(`\n❌ ${resolved.detail}`);
         process.exit(1);
       }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -11,7 +11,7 @@ import { disable } from "./commands/disable.js";
 import { enable } from "./commands/enable.js";
 import { remove, removeLibrary } from "./commands/remove.js";
 import { verify, formatVerify } from "./commands/verify.js";
-import { init } from "./commands/init.js";
+import { init, resolveInitTarget } from "./commands/init.js";
 import { upgradeCore, formatUpgrade } from "./commands/upgrade-core.js";
 import { selfUpdate, formatSelfUpdate, checkSelfUpdate, formatSelfUpdateCheck } from "./commands/self-update.js";
 import {
@@ -573,7 +573,7 @@ program
   });
 
 program
-  .command("init <name>")
+  .command("init [name]")
   .description("Scaffold a new skill, tool, agent, or prompt repo")
   .option("-d, --dir <path>", "Target directory")
   .option("-a, --author <name>", "Author GitHub username")
@@ -583,7 +583,7 @@ program
   )
   .action(
     async (
-      name: string,
+      name: string | undefined,
       opts: { dir?: string; author?: string; type?: string }
     ) => {
       const validTypes = ["skill", "tool", "agent", "prompt", "pipeline"] as const;
@@ -601,17 +601,20 @@ program
         process.exit(1);
       }
 
-      // Sanitize name — prevent path traversal
-      if (/[\/\\]|\.\./.test(name)) {
-        console.error(`\n❌ Invalid name "${name}". Name must not contain path separators or "..".`);
+      // arc#107 — resolve name + targetDir with init-in-place semantics.
+      // See `resolveInitTarget` for the matrix; logic lives there as a pure
+      // function so it's unit-testable.
+      const resolved = resolveInitTarget({
+        argName: name,
+        cwd: process.cwd(),
+        dirOverride: opts.dir,
+      });
+      if (!resolved) {
+        console.error(`\n❌ Invalid name. Name must not contain path separators or "..".`);
         process.exit(1);
       }
 
-      const prefix = `arc-${artifactType}`;
-      const targetDir =
-        opts.dir ??
-        `./${prefix}-${name.replace(/^_/, "").toLowerCase()}`;
-      const result = await init(targetDir, name, opts.author, artifactType);
+      const result = await init(resolved.targetDir, resolved.name, opts.author, artifactType);
 
       if (result.success) {
         console.log(`\n✅ Scaffolded ${artifactType} at ${result.path}`);
--- a/src/commands/init-resolve.ts
+++ b/src/commands/init-resolve.ts
@@ -1,0 +1,80 @@
+import { basename, join } from "path";
+import { platform } from "os";
+
+/**
+ * Resolve `arc init`'s `[name]` arg + cwd into a `{name, targetDir}` tuple
+ * — pure function so it can be unit-tested without spawning the CLI.
+ *
+ * arc#107 semantics:
+ *   - argless OR `.` OR `<name>` matching basename(cwd) → scaffold in cwd
+ *   - `<name>` different from basename(cwd) → ./<name>/ (no `arc-<type>-` prefix)
+ *   - explicit `dirOverride` (CLI `--dir`) always wins for targetDir
+ *
+ * Discriminated union return: `{ok: true, ...}` on success,
+ * `{ok: false, reason, detail}` on validation failure. Discriminate
+ * via `r.ok` (explicit boolean per sage P148 cycle 3 — implicit
+ * field-absence discriminators are fragile to type drift).
+ */
+export type ResolvedInitTarget =
+  | { ok: true; name: string; targetDir: string }
+  | { ok: false; reason: "invalid-name" | "invalid-dir"; detail: string };
+
+export function resolveInitTarget(opts: {
+  argName?: string;
+  cwd: string;
+  dirOverride?: string;
+  /**
+   * Override the platform for basename case-folding (test isolation).
+   * Production callers omit this and `os.platform()` decides. Sage P148
+   * cycle 5 — macOS / Windows are case-insensitive by default, so
+   * `arc init Foo` in `/x/foo` should match in-place.
+   */
+  platformOverride?: NodeJS.Platform;
+}): ResolvedInitTarget {
+  const cwdBasename = basename(opts.cwd);
+  const arg = opts.argName?.trim();
+  // Case-insensitive match on darwin / win32 (their default filesystems
+  // are case-insensitive); strict on linux + others. The actual `name`
+  // returned uses the cwd basename's casing when matched, so the
+  // manifest reflects what the filesystem actually shows.
+  const plat = opts.platformOverride ?? platform();
+  const caseInsensitive = plat === "darwin" || plat === "win32";
+  const nameMatches =
+    arg !== undefined &&
+    arg !== "" &&
+    arg !== "." &&
+    (caseInsensitive
+      ? arg.toLowerCase() === cwdBasename.toLowerCase()
+      : arg === cwdBasename);
+  // `inCwd` decides whether we scaffold in cwd (argless / `.` / matching
+  // name) or in a new subdir. Compute once.
+  const inCwd = !arg || arg === "." || nameMatches;
+
+  const name = inCwd ? cwdBasename : arg!;
+
+  if (!name || /[\/\\]|\.\./.test(name)) {
+    return {
+      ok: false,
+      reason: "invalid-name",
+      detail: `"${name}" is not a valid package name (no path separators, no "..", non-empty).`,
+    };
+  }
+
+  // Sage P148 security: validate `--dir` parity with name. Prevent
+  // path-traversal-into-shadow scenarios where a wrapper passes
+  // untrusted input through `--dir`.
+  if (opts.dirOverride !== undefined) {
+    if (opts.dirOverride === "" || /\.\.(\/|\\|$)/.test(opts.dirOverride)) {
+      return {
+        ok: false,
+        reason: "invalid-dir",
+        detail: `"${opts.dirOverride}" is not a valid --dir target (no "..", non-empty).`,
+      };
+    }
+  }
+
+  const targetDir =
+    opts.dirOverride ?? (inCwd ? opts.cwd : join(opts.cwd, name));
+
+  return { ok: true, name, targetDir };
+}
--- a/src/commands/init-scaffold.ts
+++ b/src/commands/init-scaffold.ts
@@ -1,0 +1,460 @@
+export type ArtifactInitType = "skill" | "tool" | "agent" | "prompt" | "pipeline";
+
+/**
+ * A single file `init()` will create. The list returned by
+ * {@link scaffoldEntriesFor} is the SOLE source of truth for what
+ * `init()` writes — pre-flight overwrite check, mkdir-of-parents, and
+ * the actual write loop all iterate the same array. Previously a
+ * paths-only array sat parallel to inline `Bun.write` calls, requiring
+ * a drift-guard test to catch skew. Sage P148 cycles 4 / 5 / 6 flagged
+ * the parallel sources repeatedly; cycle 7's refactor consolidated
+ * them into this shape.
+ */
+export interface ScaffoldEntry {
+  /** Path relative to the scaffold's target directory. */
+  path: string;
+  /** File contents to write verbatim. */
+  content: string;
+}
+
+/**
+ * Enumerate every file `init()` will create for a given artifact type,
+ * with contents. Single source of truth — pre-flight check + writes
+ * both iterate this array.
+ */
+export function scaffoldEntriesFor(
+  type: ArtifactInitType,
+  name: string,
+  author?: string,
+): ScaffoldEntry[] {
+  const lowerName = name.replace(/^_/, "").toLowerCase();
+  const authorName = author ?? "username";
+  const prefix = `arc-${type}`;
+
+  const entries: ScaffoldEntry[] = [
+    { path: "arc-manifest.yaml", content: buildManifest(type, name, lowerName, authorName) },
+    { path: "package.json", content: buildPackageJson(type, name, lowerName, prefix) },
+    { path: "README.md", content: buildReadme(type, name, lowerName, prefix) },
+    { path: ".gitignore", content: GITIGNORE },
+  ];
+
+  switch (type) {
+    case "tool":
+      entries.push({ path: `${lowerName}.ts`, content: buildToolEntry(name) });
+      break;
+    case "skill":
+      entries.push(
+        { path: "skill/SKILL.md", content: buildSkillMd(name) },
+        { path: "skill/workflows/Main.md", content: SKILL_WORKFLOW_MAIN },
+      );
+      break;
+    case "agent":
+      entries.push({ path: `agent/${lowerName}.md`, content: buildAgentMd(name) });
+      break;
+    case "prompt":
+      entries.push({ path: `prompt/${lowerName}.md`, content: buildPromptMd(name) });
+      break;
+    case "pipeline":
+      entries.push(
+        { path: "pipeline.yaml", content: buildPipelineYaml(name) },
+        { path: "A_EXAMPLE/action.json", content: PIPELINE_ACTION_JSON },
+        { path: "A_EXAMPLE/action.ts", content: PIPELINE_ACTION_TS },
+      );
+      break;
+  }
+
+  return entries;
+}
+
+// ── Content builders ────────────────────────────────────────────────────
+//
+// Pure functions; no side effects. Each returns the file body verbatim.
+
+function buildManifest(
+  type: ArtifactInitType,
+  name: string,
+  lowerName: string,
+  authorName: string,
+): string {
+  const header = `# arc-manifest.yaml — capability declaration
+schema: arc/v1
+name: ${name}
+version: 1.0.0
+type: ${type}
+tier: custom
+
+author:
+  name: ${authorName}
+  github: ${authorName}
+
+`;
+  switch (type) {
+    case "tool":
+      return header + `provides:
+  cli:
+    - command: "bun ${lowerName}.ts"
+      name: "${lowerName}"
+  # hooks:
+  #   - event: PostToolUse
+  #     command: "\${PAI_DIR}/hooks/MyHook.hook.ts"
+
+depends_on:
+  tools:
+    - name: bun
+      version: ">=1.0.0"
+
+capabilities:
+  filesystem:
+    read: []
+    write: []
+  network: []
+  bash:
+    allowed: false
+  secrets: []
+`;
+    case "skill":
+      return header + `provides:
+  skill:
+    - trigger: "${lowerName}"
+  # cli:
+  #   - command: "bun src/tool.ts"
+  # hooks:
+  #   - event: PostToolUse
+  #     command: "\${PAI_DIR}/hooks/MyHook.hook.ts"
+
+depends_on:
+  tools:
+    - name: bun
+      version: ">=1.0.0"
+
+capabilities:
+  filesystem:
+    read: []
+    write: []
+  network: []
+  bash:
+    allowed: false
+  secrets: []
+`;
+    case "agent":
+      return header + `capabilities:
+  filesystem:
+    read:
+      - agent/
+  network: []
+  bash:
+    allowed: false
+  secrets: []
+`;
+    case "pipeline":
+      return header + `capabilities:
+  filesystem:
+    read: []
+    write: []
+  network: []
+  bash:
+    allowed: true
+    restricted_to:
+      - bun
+  secrets: []
+`;
+    case "prompt":
+      return header + `capabilities:
+  filesystem:
+    read: []
+    write: []
+  network: []
+  bash:
+    allowed: false
+  secrets: []
+`;
+  }
+}
+
+function buildPackageJson(
+  type: ArtifactInitType,
+  name: string,
+  lowerName: string,
+  prefix: string,
+): string {
+  const typeLabel = type.charAt(0).toUpperCase() + type.slice(1);
+  return JSON.stringify(
+    {
+      name: `${prefix}-${lowerName}`,
+      version: "1.0.0",
+      description: `arc ${name} ${typeLabel}`,
+      type: "module",
+      scripts: { test: "bun test" },
+      dependencies: {},
+    },
+    null,
+    2,
+  ) + "\n";
+}
+
+function buildReadme(
+  type: ArtifactInitType,
+  name: string,
+  lowerName: string,
+  prefix: string,
+): string {
+  const pkg = `${prefix}-${lowerName}`;
+  switch (type) {
+    case "tool":
+      return `# ${name}
+
+arc tool — [brief description].
+
+## Setup
+
+\`\`\`bash
+arc install ${pkg}
+\`\`\`
+
+## Manual Setup
+
+\`\`\`bash
+git clone [repo-url] ~/Developer/${pkg}/
+cd ~/Developer/${pkg}/
+bun install
+\`\`\`
+
+## License
+
+MIT
+`;
+    case "skill":
+      return `# ${name}
+
+arc skill — [brief description].
+
+## Setup
+
+\`\`\`bash
+git clone [repo-url] ~/Developer/${pkg}/
+cd ~/Developer/${pkg}/
+bun install
+\`\`\`
+
+## Integration
+
+\`\`\`bash
+ln -sfn ~/Developer/${pkg}/skill ~/.claude/skills/${name}
+\`\`\`
+
+## License
+
+MIT
+`;
+    case "agent":
+      return `# ${name}
+
+arc agent — [brief description].
+
+## Setup
+
+\`\`\`bash
+git clone [repo-url] ~/Developer/${pkg}/
+\`\`\`
+
+## Integration
+
+\`\`\`bash
+ln -sfn ~/Developer/${pkg}/agent ~/.claude/agents/${name}
+\`\`\`
+
+## License
+
+MIT
+`;
+    case "pipeline":
+      return `# ${name}
+
+arc pipeline — [brief description].
+
+## Setup
+
+\`\`\`bash
+arc install ${pkg}
+\`\`\`
+
+## Manual Setup
+
+\`\`\`bash
+git clone [repo-url] ~/Developer/${pkg}/
+cd ~/Developer/${pkg}/
+bun install
+\`\`\`
+
+## License
+
+MIT
+`;
+    case "prompt":
+      return `# ${name}
+
+arc prompt — [brief description].
+
+## Setup
+
+\`\`\`bash
+git clone [repo-url] ~/Developer/${pkg}/
+\`\`\`
+
+## Integration
+
+\`\`\`bash
+ln -sfn ~/Developer/${pkg}/prompt ~/.claude/prompts/${name}
+\`\`\`
+
+## License
+
+MIT
+`;
+  }
+}
+
+function buildToolEntry(name: string): string {
+  return `#!/usr/bin/env bun
+
+/**
+ * ${name} — arc tool
+ */
+
+console.log("${name} tool");
+`;
+}
+
+function buildSkillMd(name: string): string {
+  return `---
+name: ${name}
+description: |
+  [Describe what this skill does]
+
+  USE WHEN user says "[trigger phrase]"
+---
+
+# ${name}
+
+[Instructions for the AI agent]
+
+## Workflow Routing
+
+| Trigger | Workflow |
+|---------|----------|
+| [trigger] | \`Workflows/Main.md\` |
+`;
+}
+
+function buildAgentMd(name: string): string {
+  return `---
+name: ${name}
+description: "[Describe what this agent does]"
+model: sonnet
+
+persona:
+  name: "${name}"
+  title: "[Agent title or role]"
+  background: |
+    [Describe the agent's background, expertise, and personality]
+---
+
+# ${name}
+
+[Detailed instructions for how this agent should behave]
+
+## Capabilities
+
+- [Capability 1]
+- [Capability 2]
+
+## Constraints
+
+- [Constraint 1]
+- [Constraint 2]
+`;
+}
+
+function buildPromptMd(name: string): string {
+  return `---
+name: ${name}
+description: "[Describe what this prompt does]"
+version: 1.0.0
+---
+
+# ${name}
+
+## System
+
+[System-level instructions or context for the model]
+
+## User Template
+
+[The prompt template. Use {{variable}} for placeholders.]
+
+## Variables
+
+| Variable | Description | Required |
+|----------|-------------|----------|
+| {{variable}} | [Description] | yes |
+
+## Example
+
+**Input:**
+\`\`\`
+[Example input]
+\`\`\`
+
+**Output:**
+\`\`\`
+[Expected output]
+\`\`\`
+`;
+}
+
+function buildPipelineYaml(name: string): string {
+  return `name: ${name}
+description: "[Describe what this pipeline does]"
+version: 1.0.0
+
+actions:
+  - name: A_EXAMPLE
+    description: "Example action"
+`;
+}
+
+const SKILL_WORKFLOW_MAIN = `# Main Workflow
+
+## Steps
+
+1. [Step 1]
+2. [Step 2]
+`;
+
+const PIPELINE_ACTION_JSON = JSON.stringify(
+  {
+    name: "A_EXAMPLE",
+    description: "Example action — replace with your logic",
+    inputs: { data: { type: "string", description: "Input data" } },
+    outputs: { result: { type: "string", description: "Output result" } },
+  },
+  null,
+  2,
+) + "\n";
+
+const PIPELINE_ACTION_TS = `#!/usr/bin/env bun
+
+/**
+ * A_EXAMPLE — example pipeline action
+ */
+
+const input = JSON.parse(process.argv[2] ?? "{}");
+console.log(JSON.stringify({ result: \`Processed: \${input.data}\` }));
+`;
+
+const GITIGNORE = `node_modules/
+.env
+*.env
+secrets/
+cache/
+`;
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,5 +1,5 @@
 import { basename, join } from "path";
-import { existsSync } from "fs";
+import { existsSync, lstatSync } from "fs";
 import { mkdir } from "fs/promises";
 
 export type ArtifactInitType = "skill" | "tool" | "agent" | "prompt" | "pipeline";
@@ -95,6 +95,16 @@ export async function init(
   author?: string,
   type: ArtifactInitType = "skill"
 ): Promise<InitResult> {
+  // Sage P148 cycle 2: refuse when targetDir is an existing non-directory
+  // (regular file, symlink to file). `mkdir({recursive: true})` would
+  // throw `EEXIST` mid-scaffold, partially written. Surface a clean
+  // error first.
+  if (existsSync(targetDir) && !lstatSync(targetDir).isDirectory()) {
+    return {
+      success: false,
+      error: `${targetDir} exists and is not a directory`,
+    };
+  }
   // Refuse only if a manifest already lives at the target — the
   // unambiguous signal that the directory is already an arc package.
   // Other files (a fresh git clone with only .git/, an empty package.json
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,153 +1,14 @@
-import { basename, dirname, join } from "path";
+import { dirname, join } from "path";
 import { existsSync, lstatSync, statSync } from "fs";
 import { mkdir } from "fs/promises";
-import { platform } from "os";
+import { scaffoldEntriesFor, type ArtifactInitType } from "./init-scaffold.js";
 
-export type ArtifactInitType = "skill" | "tool" | "agent" | "prompt" | "pipeline";
-
-/**
- * A single file the scaffold will create. The list returned by
- * {@link scaffoldEntriesFor} is the SOLE source of truth for what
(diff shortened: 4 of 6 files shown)


jcfischer (commented):
## Sage code review — commented

6 finding(s): 2 important, 2 suggestion, 2 nit.


### CodeQuality
The PR is well-structured with clean separation of concerns — extracting `resolveInitTarget` as a pure function for testability is idiomatic. Two minor quality issues: `lstatSync` incorrectly rejects symlinks to directories despite the downstream `mkdir` handling them fine, and the JSDoc for `resolveInitTarget` inaccurately claims it returns `null` when it returns a `ResolveInitFailure` discriminated union.

- **[suggestion]** `src/commands/init.ts:101` — **`lstatSync` rejects symlinks to directories unnecessarily**
  The guard `!lstatSync(targetDir).isDirectory()` uses `lstatSync`, which does not follow symlinks. A symlink pointing to a valid directory is rejected with "exists and is not a directory", even though `mkdir({recursive: true})` would follow the symlink and succeed on the target directory. The comment on line 100 says "symlink to file" — but the code also catches symlinks to directories. `statSync` (which follows symlinks) matches the intended behavior: reject only regular files and broken symlinks, not symlinks to directories.
  
  Suggestion:

  ```
  Replace `lstatSync(targetDir).isDirectory()` with `statSync(targetDir).isDirectory()`, changing the import from `lstatSync` to `statSync`.
  ```

- **[nit]** `src/commands/init.ts:17` — **Docstring says "Returns `null`" but function returns a typed failure object**
  The JSDoc for `resolveInitTarget` states "Returns `null` when the resolved name OR a `--dir` override would be invalid", but the function signature is `ResolvedInitTarget | ResolveInitFailure` — it never returns `null`. This is misleading to readers relying on the docstring to understand the return type.
  
  Suggestion:

  ```
  Change to: "Returns a `ResolveInitFailure` when the resolved name OR a `--dir` override would be invalid…"
  ```

### Security
The PR adds solid --dir traversal validation and an lstatSync guard, both security improvements. However, the new init-in-place mode silently overwrites any existing files at scaffold paths (README.md, package.json, .gitignore, skill/SKILL.md, etc.) via Bun.write, with only arc-manifest.yaml guarded. The PR description's claim that it 'layers cleanly without overwriting' is inaccurate and masks a data-loss risk in wrapper/automation contexts.

- **[important]** `src/commands/init.ts:95` — **Silent file overwrite in init-in-place mode — only arc-manifest.yaml is guarded**
  The new init() logic allows scaffolding into existing directories, guarding only against arc-manifest.yaml pre-existence. All other Bun.write calls (README.md, package.json, .gitignore, skill/SKILL.md, skill/workflows/Main.md, and type-specific scaffold files) silently overwrite any existing file at those paths. A user who runs `arc init` in a non-arc project directory, or a wrapper script that passes --dir pointing at an existing project, will lose custom README, package.json dependencies/scripts, and other content replaced by arc templates. This is a data-integrity concern at the tool's primary trust boundary (user intent via CLI invocation).
  
  Suggestion:

  ```
  Either (a) skip file writes when the target already exists (true 'layer cleanly'), or (b) emit a warning and prompt for confirmation before overwriting existing files, or (c) update the PR description to accurately reflect that standard scaffold files will be replaced.
  ```

### Architecture
The PR extracts a pure `resolveInitTarget` function from inline CLI logic into the commands layer — an architecturally sound move that thins the CLI, makes resolution independently testable, and keeps the command module cohesive. Two structural concerns: the discriminated union uses an implicit (field-absence) discriminator, and the test filename embeds an issue number rather than describing the behavior under test.

- **[important]** `src/commands/init.ts:27` — **Implicit discriminated union — relies on field absence for type narrowing**
  `ResolvedInitTarget` and `ResolveInitFailure` are discriminated by checking `'reason' in resolved`. This is fragile: if `ResolvedInitTarget` ever gains a `reason` field, the narrowing silently breaks with no compile error. A proper tagged union (e.g., `ok: true` vs `ok: false`, or a shared `kind` discriminant) makes the intent explicit and compiler-enforceable. Forward-compatibility risk — the current pattern couples type safety to a field name that only one side owns.
  
  Suggestion:

  ```
  Add an explicit discriminant field shared by both types, e.g., `kind: 'ok'` / `kind: 'error'`, or `success: true` / `success: false`. Narrow on that instead of `'reason' in resolved`.
  ```

- **[suggestion]** `test/unit/init-resolve-107.test.ts:1` — **Test filename embeds issue number rather than describing behavior**
  The file `init-resolve-107.test.ts` ties test organization to an issue tracker artifact. Over time, as more issue-anchored test files accumulate, mapping behavior-to-file requires git archaeology. Naming by the behavior under test (`init-target-resolution.test.ts`) scales better and keeps the test directory self-documenting.
  
  Suggestion:

  ```
  Rename to `test/unit/init-target-resolution.test.ts` or co-locate these tests in `test/commands/init.test.ts` alongside the existing init integration tests (the resolve function is pure and doesn't strictly need a separate file).
  ```

### Performance
This PR modifies the one-shot `arc init` CLI scaffolding path. No hot-path loops, database queries, caching, or network access are introduced. The sync fs calls (`existsSync`, `lstatSync`) in the async `init()` function are trivial checks before a one-time directory creation — no measurable degradation given the command is run once per project initialization.

_No findings._

### Maintainability
The PR is clean from a maintainability standpoint. The new `resolveInitTarget` pure function is well-scoped, testable, and appropriately colocated with `init`. One minor DRY observation: the subexpression `!arg || arg === "."` is evaluated twice within the same function body, which is a nit but not worth holding up merge.

- **[nit]** `src/commands/init.ts:43` — **Duplicate subexpression `!arg || arg === "."` in resolveInitTarget**
  The condition `!arg || arg === "."` appears on line 43 (to choose `name`) and again as a subexpression on line 53 (to compute `inCwd`). The two evaluations can't drift apart today since `arg` is derived from a const `trim()` and never reassigned, but a future edit that mutates `arg` between them would silently break the invariant. Extracting a single `const isArglessOrDot` makes the coupling explicit.
  
  Suggestion:

  ```
  Extract `const isArglessOrDot = !arg || arg === "."` immediately after the `const arg = ...` line, then reuse it in both the `name` ternary and the `inCwd` expression.
  ```


---
_Posted by Sage on pi.dev substrate._

jcfischer (commented):
## Sage code review — commented

5 finding(s): 1 important, 3 suggestion, 1 nit.


### CodeQuality
The PR introduces a clean pure-function resolver (`resolveInitTarget`) with well-structured discriminated-union error handling and thorough tests. The init-in-place semantics are correctly implemented. Two real issues surfaced: (1) `init()` throws an unhandled promise rejection when `targetDir` is a broken symlink, contradicting the PR's own documentation that claims broken symlinks are handled cleanly via the returned `InitResult`; (2) the in-place detection uses case-sensitive comparison, which will surprise users on case-insensitive filesystems (macOS default, Windows) when the cwd basename and arg differ only in case.

- **[important]** `src/commands/init.ts:95` — **`init()` throws unhandled on broken-symlink target, contradicting documented behavior**
  `existsSync()` returns `false` for broken symlinks, so the early `!isDirectory()` guard is skipped. Control falls through to `mkdir(targetDir, {recursive: true})`, which throws `ENOENT` on a broken symlink. This is an unhandled promise rejection — the function returns `Promise<InitResult>` but propagates the exception instead. The PR's own code comment (line 113–115) claims broken symlinks "fail fast with a clean error before any partial state lands," but the error is an uncaught throw, not a returned `InitResult`.
  
  Suggestion:

  ```
  Add an explicit broken-symlink check before mkdir, e.g. `if (!existsSync(targetDir) && lstatSync(targetDir).isSymbolicLink()) { return { success: false, error: `...` }; }` using `lstatSync` (which does not follow symlinks), or wrap `mkdir` in try/catch and convert to an `InitResult` error.
  ```

- **[suggestion]** `src/commands/init.ts:70` — **Case-sensitive `cwdBasename` comparison breaks init-in-place on case-insensitive filesystems**
  `inCwd` uses `arg === cwdBasename`, which is case-sensitive. On macOS (default APFS is case-insensitive) and Windows, a user who creates `mkdir MySkill` and runs `arc init myskill` from inside it will get a nested `./myskill/` directory rather than init-in-place, because `"myskill" !== "MySkill"`. This contradicts the operator expectation set by the resolution matrix in the PR description.
  
  Suggestion:

  ```
  Consider `arg.toLowerCase() === cwdBasename.toLowerCase()` for the matching branch, or alternatively document this as intentional strict matching.
  ```

### Security
Security review of PR #148 (arc init in-place scaffold). The PR adds path-traversal validation for both the [name] argument and --dir option, plus a pre-flight file-overwrite check via scaffoldFilesFor. No concrete attack paths found; the changes are net security-positive — the prior code had no path-traversal guard on name at all, and the new code catches separators and .. in both inputs. TOCTOU between file-exists check and write is noted but irrelevant for a local CLI tool.

_No findings._

### Architecture
The PR makes a sound architectural tradeoff — extracting `resolveInitTarget` as a pure function from `cli.ts` into `commands/init.ts` for testability — but this fractures the command boundary: `cli.ts` now imports two exports from one command module and knows about resolution internals that could have stayed encapsulated inside `init()`. The scaffold-file enumeration (`scaffoldFilesFor`) is a well-guarded maintenance hotspot; the drift-guard test is the right mitigation but the dual-maintenance risk (list + writes) remains a structural cost.

- **[suggestion]** `src/commands/init.ts:89` — **resolveInitTarget exported as public command-surface breaks init encapsulation**
  `cli.ts` now imports both `init` and `resolveInitTarget` from the same command module, giving the CLI entry point knowledge of a two-phase resolution-then-execute protocol that could live entirely inside `init()`. The stated justification — 'pure function so it can be unit-tested' — is valid, but the same testability could be achieved by having `init()` accept raw args internally call `resolveInitTarget` (unexported), and return a discriminated `InitResult` with a `reason` field for validation failures. The current approach expands the public surface of `commands/init.ts` by two symbols (`resolveInitTarget`, `ResolvedInitTarget`) whose only consumer is the adjacent `cli.ts` caller.
  
  Suggestion:

  ```
  Consider making `resolveInitTarget` module-private and having `init()` accept `{ argName?, cwd, dirOverride? }` directly. The CLI action would then call a single function and pattern-match on `InitResult.success` + `InitResult.error`, matching the pattern already used post-init. The pure-function unit tests can still exercise resolution indirectly through `init()` or by extracting `resolveInitTarget` into a separate internal-only module (e.g. `src/commands/init-resolve.ts`) that is not re-exported.
  ```

- **[suggestion]** `src/commands/init.ts:14` — **scaffoldFilesFor creates a dual-maintenance hazard between file list and write logic**
  `scaffoldFilesFor` enumerates the full file set for each artifact type, but the actual `Bun.write` calls in `init()` (below line ~110) are independently maintained. Adding a new scaffold file requires coordinated edits in two places within the same function body. The PR correctly includes a drift-guard test that catches mismatches, but the root cause is structural: the file list is declarative while the write logic is imperative, and nothing statically binds them together. Over time, this pattern invites silent drift where a write is added without updating the list (or vice versa), relying entirely on the test to catch it.
  
  Suggestion:

  ```
  Consider driving the scaffold writes from the `scaffoldFilesFor` list itself — generate content per file path rather than maintaining separate `Bun.write` blocks. A simple dispatch map (`path → content-generator`) keyed by the same relative paths would make the list the single source of truth, eliminating the dual-maintenance problem entirely.
  ```

- **[nit]** `test/unit/init-resolve-107.test.ts:1` — **Issue-numbered test file name encourages sprawl**
  The test file is named `init-resolve-107.test.ts`, tying it to a specific issue rather than a tested concern. If future resolution changes span issues #107, #215, and #340, this naming pattern leads to either a single catch-all file (defeating the naming convention) or multiple issue-numbered files that duplicate setup and make it hard to understand what behaviors are covered. The test suite already has `test/commands/init.test.ts` covering the init command; a concern-named file like `test/unit/init-resolve.test.ts` would scale better.
  
  Suggestion:

  ```
  Rename to `init-resolve.test.ts` (dropping the `-107` suffix) and add a comment referencing #107 for provenance. Keep the issue reference in the test description strings where it already appears.
  ```

### Performance
No performance concerns. This PR modifies a one-shot CLI init command — all operations are O(1) string manipulation or trivial filesystem checks on <10 paths. There are no hot paths, network calls, database queries, caches, or unbounded growth patterns affected by this change.

_No findings._

### Maintainability
This PR improves maintainability end-to-end. The extraction of `resolveInitTarget` (pure, testable, discriminated union) and `scaffoldFilesFor` (single source of truth for file enumeration) reduces the CLI action handler to a thin orchestration layer. No duplication, no deep nesting, no dead exports, and parameter counts are well within bounds. The drift-guard test is a thoughtful touch that protects against future write-vs-list skew.

_No findings._


---
_Posted by Sage on pi.dev substrate._

jcfischer (commented):
## Sage code review — commented

7 finding(s): 2 important, 2 suggestion, 3 nit.


### CodeQuality
The `resolveInitTarget` pure function and the reworked `init()` pre-flight checks are correctly implemented with thorough edge-case coverage (symlinks, broken symlinks, case-insensitive matching on macOS/Windows, path-traversal rejection). The production code is idiomatic, clear, and well-tested. One minor cross-platform concern exists in a test helper, but no correctness or clarity issues in the changed source.

- **[nit]** `test/commands/init.test.ts:160` — **Drift-guard test walk() hardcodes `/` separator**
  The `walk()` helper constructs relative paths with template literal `${base}/${entry.name}`, always using `/`. Meanwhile `result.files` is populated by the `init()` function, which likely uses `join()` and thus the platform-native separator. On Windows this would produce mismatched separators (e.g. `skill\SKILL.md` vs `skill/SKILL.md`), causing `.toEqual()` to fail even when the file sets are identical. The test is otherwise well-designed for catching list-vs-write drift.
  
  Suggestion:

  ```
  Either normalize both sets to `/` before comparison, or use `join()` consistently in `walk()` so the separators match.
  ```

### Security
This PR is security-positive: it adds path-traversal validation on both the `name` argument and `--dir` override, a pre-flight check that refuses to overwrite any existing scaffold target files (preventing accidental data loss), proper symlink handling (rejecting broken links and non-directory files), and a drift-guard test that keeps the file-enumeration list in lockstep with actual writes. No injection surfaces, secrets, auth, or crypto concerns are present in this diff.

_No findings._

### Architecture
The PR extracts a clean `resolveInitTarget` pure function (good layering — CLI delegates to command module, dependency direction is correct) and introduces `scaffoldFilesFor` as a central file manifest. The main architectural tension is that `scaffoldFilesFor` and the actual `Bun.write` calls inside `init()` are maintained as two independent sources of truth — a structural coupling that the drift-guard test defends against but doesn't eliminate. The test organization (`test/unit/` vs `test/commands/`) is well-reasoned.

- **[important]** `src/commands/init.ts:23` — **Dual source-of-truth: `scaffoldFilesFor` and `init()` writes maintained independently**
  `scaffoldFilesFor` declares the set of files a scaffold will write, but `init()` contains its own inline `Bun.write` calls and `files.push()` tracking that are not driven from this list. If a future contributor adds a `Bun.write` for a new file without updating `scaffoldFilesFor`, the pre-flight overwrite check silently misses it and operator content can be clobbered. The drift-guard test in `test/commands/init.test.ts` catches this at test time, but the structural duplication invites the defect. A single data structure (e.g., a `Map<relativePath, contentFactory>`) that both the pre-flight check and the write loop consume would eliminate the coupling.
  
  Suggestion:

  ```
  Refactor `init()` so the write section iterates over `scaffoldFilesFor` (or a richer `Map<relPath, contentGenerator>`) rather than issuing independent `Bun.write` calls. This makes the file list the single source of truth, and the drift-guard test becomes redundant rather than load-bearing.
  ```

- **[suggestion]** `src/commands/init.ts:96` — **`platformOverride` testing seam pollutes the public API of a pure function**
  `resolveInitTarget` is documented as a pure function, but its `opts` interface exposes `platformOverride?: NodeJS.Platform` — a parameter that exists solely for test isolation. Production callers must either omit it (defaulting to `os.platform()`, an impure call) or pass it explicitly. This leaks test infrastructure into the public API surface. A cleaner separation would accept a `caseSensitive: boolean` parameter computed by the CLI layer (which already has access to `process.platform`), keeping the resolver truly pure and the platform concern in the orchestrator.
  
  Suggestion:

  ```
  Replace `platformOverride` with `caseSensitive?: boolean`. Compute `caseSensitive = process.platform === 'darwin' || process.platform === 'win32'` in `cli.ts` before calling `resolveInitTarget`, and pass it explicitly. This eliminates both the `os` import and the testing seam from the command module.
  ```

- **[nit]** `src/commands/init.ts:131` — **`init()` is accumulating four distinct responsibilities**
  The function now handles: (1) target path validation (file/dir/symlink/broken-symlink probes), (2) pre-flight overwrite checks against `scaffoldFilesFor`, (3) `mkdir` with recursive creation, and (4) scaffold writes. These are still cohesive (all relate to "scaffold into a target"), but the function is growing toward ~150 lines with nested try/catch and switch logic. Extracting a `validateTarget(targetDir): ValidationResult` helper would improve readability and make each responsibility independently testable without mocking the filesystem for the entire init flow.
  
  Suggestion:

  ```
  Extract the lstat/stat/symlink validation block (lines ~139–181 in the new code) into a private `validateTargetPath(targetDir: string): { ok: true } | { ok: false; error: string }` function. This keeps `init()` focused on orchestration and makes the validation logic unit-testable without scaffolding a full init.
  ```

### Performance
This PR touches only CLI init-time code — a pure resolver function called once and a one-shot scaffold routine. No hot-path loops, no database access, no async I/O bottlenecks, no closures in render paths, and no unbounded caches or queues. The sync filesystem calls (`existsSync`, `lstatSync`, `statSync`) inside `init()` operate on at most ~6 files and run once at CLI invocation, not in any request-serving or interval path. No performance degradation is inferable.

_No findings._

### Maintainability
The PR extracts a clean `resolveInitTarget` pure function — a maintainability win. However, `scaffoldFilesFor` and the hardcoded `Bun.write` calls in `init()` are two independent lists defining the same file set; they must be kept in sync manually, and the drift-guard test is a safety net rather than a structural fix. The `init()` function has also grown large (~140 lines) with extractable phases.

- **[important]** `src/commands/init.ts:19` — **`scaffoldFilesFor` and write code are two independent sources of truth for scaffold files**
  `scaffoldFilesFor` enumerates files for the pre-flight overwrite check, while the file-writing section of `init()` has its own hardcoded series of `Bun.write` calls. A future contributor who adds a new scaffold file must update both places. Updating only the write path means the pre-flight check silently misses the new file and operator content could be clobbered. Updating only the list means init spuriously refuses on a file it will never write. The drift-guard test in `init.test.ts` catches some divergence scenarios but is an after-the-fact safety net rather than a structural solution.
  
  Suggestion:

  ```
  Drive both the pre-flight check and the actual writes from a single data structure (e.g., a `ScaffoldFile[]` array with `{ relPath, contentGenerator }` entries). This eliminates the duplication and makes adding a new file a single-place change.
  ```

- **[suggestion]** `src/commands/init.ts:131` — **`init()` has grown large with extractable path-validation phase**
  The function now spans ~140 lines with three distinct sequential phases: (1) target-path validation including lstat/symlink/file detection (~35 lines, nesting depth 3), (2) pre-flight overwrite check (~15 lines), and (3) scaffolding (~80 lines). The path-validation block is a self-contained concern with its own error-return shape and no coupling to later phases beyond the targetDir. Extracting it would reduce cognitive load when reading the scaffolding logic.
  
  Suggestion:

  ```
  Extract a `validateTargetPath(targetDir: string): { ok: true } | { ok: false; error: string }` helper. While it would be a single-callsite helper, the size and nesting depth of the block outweigh the single-callsite concern in this case.
  ```

- **[nit]** `src/commands/init.ts:69` — **`platformOverride` is test-only infrastructure in the production API**
  The `platformOverride` parameter in `resolveInitTarget` exists solely so tests can inject a platform for case-sensitivity assertions. Production callers never pass it. This adds an optional parameter to the public API surface that exists only to satisfy test isolation, which is a mild abstraction leak. A future reader of the production call site may wonder why it exists.
  
  Suggestion:

  ```
  Consider wrapping `os.platform()` in a module-level variable (e.g., `let _platform = platform`) that tests can monkey-patch via a `__setPlatformForTest` export, keeping the production API clean. Alternatively, accept it as documented test infrastructure — the current pattern is common and not harmful.
  ```


---
_Posted by Sage on pi.dev substrate._

jcfischer (commented):
## Sage code review — commented

7 finding(s): 2 important, 4 suggestion, 1 nit.


### CodeQuality
The refactor is well-executed: `resolveInitTarget` is a clean, testable pure function, the single-source-of-truth `scaffoldEntriesFor` eliminates parallel-list drift, and the lstat/stat dance handles regular files, broken symlinks, and symlink-to-dir correctly. Two minor code-quality patterns worth addressing before merge.

- **[suggestion]** `src/commands/init.ts:1` — **Switch statements on ArtifactInitType lack exhaustiveness guard**
  `buildManifest`, `buildReadme`, and `scaffoldEntriesFor` each use a `switch` covering all five current union members with no `default` branch and no `satisfies never` exhaustiveness check. The functions all declare `: string` or `: ScaffoldEntry[]` return types, but if `ArtifactInitType` later expands, the unhandled case falls through to an implicit `undefined` return — which TypeScript may not catch without `noImplicitReturns`. A defensive `default: throw new Error(...)` or a `satisfies never` pattern would make future type additions fail loudly at compile time.
  
  Suggestion:

  ```
  Add a `default:` branch that throws, e.g.: `default: { const _exhaustive: never = type; throw new Error(`Unhandled type: ${_exhaustive}`); }`
  ```

- **[nit]** `src/commands/init.ts:1` — **Imported `platform` from 'os' used only with an optional test override**
  `resolveInitTarget` imports `platform` from `os` for the production case-insensitivity decision, but also accepts a `platformOverride` parameter solely for test isolation. The production path (`plat = opts.platformOverride ?? platform()`) works correctly, but the caller in `cli.ts` never passes `platformOverride` — this parameter exists only to satisfy the test suite. The interface is slightly cluttered.
  
  Suggestion:

  ```
  Consider inverting the dependency: have the production caller pass `os.platform()` explicitly via `platformOverride`, remove the `import { platform } from 'os'` inside the pure function, and drop the optional fallback. Keeps the function zero-dependency and the override path becomes the only path.
  ```

### Security
PR #148 introduces init-in-place scaffolding with solid security posture. Path traversal is defended at two trust boundaries (the `name` arg and `--dir` override) with regex-based validation. A pre-flight overwrite check prevents clobbering operator files. Symlink handling correctly distinguishes broken symlinks, regular-file-at-target, and valid symlinks-to-directories. No hardcoded secrets, no injection surfaces touched, and no crypto concerns. No findings to raise through the Security lens.

_No findings._

### Architecture
The PR refactors `init.ts` to consolidate scaffold-file enumeration into a single source of truth (`scaffoldEntriesFor`), extracts a pure `resolveInitTarget` for testable name/directory resolution, and adds pre-flight overwrite checks. The structural direction is sound — eliminating parallel-list drift and centralizing entry enumeration will reduce future bugs. However, the module now exports test-motivated internals (`scaffoldEntriesFor`, `resolveInitTarget` with `platformOverride`) that widen the public contract, and the file has grown to ~650+ lines mixing three distinct concerns (resolution, content definition for 5 artifact types, and execution).

- **[important]** `src/commands/init.ts:108` — **Test-only `platformOverride` leaks into production API surface**
  The `platformOverride?: NodeJS.Platform` option on `resolveInitTarget` exists solely for unit tests to simulate macOS/Windows case-insensitive matching. Production callers always omit it. Once this parameter is part of the exported `resolveInitTarget` options type, it becomes a public contract — external consumers may discover and depend on it, and removing it later becomes a breaking change. Public API surfaces should not carry test infrastructure.
  
  Suggestion:

  ```
  Consider two approaches: (a) pass the platform via a module-level mutable flag gated behind `process.env.NODE_ENV === 'test'`, keeping it out of the function signature; or (b) keep `platformOverride` but mark the entire `resolveInitTarget` function as `@internal` in JSDoc so external consumers don't depend on it.
  ```

- **[important]** `src/commands/init.ts:18` — **`scaffoldEntriesFor` and `ScaffoldEntry` exported for testing widen public contract**
  These are internal implementation details — the canonical list of files `init()` writes. They're exported so the drift-guard test can verify list-vs-write consistency, a legitimate architectural need. However, exporting them makes them public API: downstream consumers can import and depend on `ScaffoldEntry`'s shape and `scaffoldEntriesFor`'s return values, coupling external code to internal scaffold structure. Same concern for `ResolvedInitTarget` and `resolveInitTarget` — CLI-layer resolution logic now a public contract.
  
  Suggestion:

  ```
  Add `@internal` JSDoc tags to `scaffoldEntriesFor`, `ScaffoldEntry`, `resolveInitTarget`, and `ResolvedInitTarget`. This signals intent without breaking the test import. Alternatively, adopt a test-only barrel import pattern (e.g., `src/commands/init.internal.ts` re-exporting internals for `test/` only).
  ```

- **[suggestion]** `src/commands/init.ts:30` — **Asymmetric public API: entry enumerator is public but content builders are private**
  `scaffoldEntriesFor` is exported and returns pre-computed `ScaffoldEntry[]`. The content builder functions it calls (`buildManifest`, `buildPackageJson`, `buildReadme`, `buildSkillMd`, etc.) are module-private. A consumer who imports `scaffoldEntriesFor` to customize scaffold output (the natural use case for a public entry enumerator) must either fork the function or post-mutate the returned array — they can't recompose entries from the individual builders. This creates an API that looks extensible but isn't.
  
  Suggestion:

  ```
  Either: (a) export the content builders alongside `scaffoldEntriesFor` so consumers can assemble custom entry lists; or (b) mark `scaffoldEntriesFor` as `@internal` and keep the builders private, making it clear the function exists for internal use and testing only.
  ```

- **[suggestion]** `src/commands/init.ts:1` — **`init.ts` module mixes three separable concerns at ~650+ lines**
  The module now contains: (1) target-name resolution (`resolveInitTarget`, ~80 lines), (2) scaffold content definition for 5 artifact types (`scaffoldEntriesFor` + 8 content builders + 4 content constants, ~350 lines), and (3) scaffold execution (`init`, ~100 lines) plus types (~50 lines). Content builders alone span skill/tool/agent/prompt/pipeline. Future work on any artifact type's scaffold content or on the resolution/execution logic touches this monolithic file. The consolidation into a single source of truth is architecturally correct; the question is whether that truth lives in one file.
  
  Suggestion:

  ```
  Consider a future split: `src/commands/init/resolve.ts` for `resolveInitTarget`, `src/commands/init/content.ts` for `scaffoldEntriesFor` + builders + constants, and `src/commands/init/index.ts` for `init()` + types. Not blocking — the file is manageable now — but worth tracking as technical debt before the next artifact type is added.
  ```

### Performance
PR changes are confined to the CLI `arc init` path — a one-shot, operator-triggered scaffolding command that creates at most ~7 small files. No hot-path loops, no network/database access, no unbounded caches, and no closures in render paths exist in the diff. The sync I/O calls (`existsSync`, `lstatSync`, `statSync`) inside the `async init()` function are non-issues in a CLI context where the event loop has no concurrent work. No concrete performance degradation is introduced.

_No findings._

### Maintainability
The PR is a net maintainability win: it de-duplicates manifest header boilerplate (5 copies → 1 shared `header`), consolidates parallel file lists into a single `scaffoldEntriesFor` source of truth, and extracts name resolution into a pure testable function. The one friction point is the `init()` function's ~45-line path-validation preamble (lstat/symlink/ENOENT) with nesting depth 4, which mixes a distinct concern into the scaffold orchestration body.

- **[suggestion]** `src/commands/init.ts:190` — **Extract targetDir validation from init()**
  The `init()` function now opens with a ~45-line block handling lstat, ENOENT, regular files, broken symlinks, and symlinks-to-files (try/catch nested inside if/else, depth reaching 4). This is a distinct concern — "is this targetDir usable for scaffolding?" — separate from scaffold orchestration (pre-flight checks, mkdir, writes). A reader scanning `init()` for the scaffold logic has to wade through edge-case filesystem validation first. Extracting into a `validateTargetDir(dir: string): { ok: true } | { ok: false; error: string }` helper would flatten `init()` and make the validation independently testable (the tests already cover these cases via the full `init()` call, but the lstat/symlink/broken-symlink cases in `init.test.ts` would become simpler unit tests on the helper).
  
  Suggestion:

  ```
  Extract the lstat/symlink/ENOENT block (from `let lstat` through the closing `}` of the `if (lstat)` block) into a function like `function probeTargetDir(targetDir: string): { ok: true } | { ok: false; error: string }`. Call it at the top of `init()` and early-return on failure. This also removes the `let lstat; try { … } catch … if (lstat) …` mutable-variable-plus-nested-branch pattern that requires a reader to track state across 45 lines.
  ```


---
_Posted by Sage on pi.dev substrate._

jcfischer (commented):
## Sage code review — commented

9 finding(s): 1 important, 4 suggestion, 4 nit.


### CodeQuality
Well-executed refactor with strong correctness properties: the discriminated union in resolveInitTarget is used correctly, the single-source-of-truth entries array eliminates drift risk, and the pre-flight overwrite check + validateTargetDir cover edge cases (broken symlinks, regular files at target) cleanly. Three minor clarity/type-safety nits, none blocking.

- **[nit]** `src/commands/init.ts:33` — **`lstat` variable is implicitly `any`**
  Declaring `let lstat;` without a type annotation gives it type `any` because TypeScript cannot infer across the try/catch boundary. The subsequent `lstat.isSymbolicLink()` call receives no type-checking. The function is small enough that this is unlikely to cause a bug, but it's a gap in strict typing.
  
  Suggestion:

  ```
  Use `let lstat: ReturnType<typeof lstatSync>;` or `let lstat: fs.Stats;`
  ```

- **[nit]** `src/commands/init.ts:36` — **JSDoc describes `validateTargetDir` as a 'pure piece' but it performs filesystem I/O**
  The function comment on `init()` says 'three pure pieces split across sibling modules: validateTargetDir (probe), scaffoldEntriesFor (what to write), and a single write loop here.' Only `scaffoldEntriesFor` is pure; `validateTargetDir` calls `lstatSync`/`statSync` and the write loop calls `mkdir`/`Bun.write`. Misleading documentation.
  
  Suggestion:

  ```
  Replace 'pure pieces' with a more accurate descriptor, e.g. 'three concerns split across sibling modules.'
  ```

- **[nit]** `test/commands/init.test.ts:101` — **Test description references old function name `scaffoldFilesFor`**
  The test title says 'scaffoldFilesFor stays in sync with actual writes' but the function was renamed to `scaffoldEntriesFor`. The test body correctly uses `result.files` (which is built from the entries list), but the stale name in the description is confusing for readers tracing the code.
  
  Suggestion:

  ```
  Rename the test to 'scaffoldEntriesFor stays in sync with actual writes (drift guard)'.
  ```

### Security
This PR refactors `arc init` for in-place scaffolding and drops the `arc-skill-` prefix. From a security perspective, the changes are solid: path-traversal validation is centralized in `resolveInitTarget` covering both the `[name]` arg and the `--dir` override, the pre-flight overwrite check in the refactored `init()` prevents accidental clobbering of operator content, and `validateTargetDir` correctly handles symlinks, broken symlinks, and non-directory files. No concrete attack paths or security invariant violations were found.

_No findings._

### Architecture
The PR splits a monolithic init.ts into three modules with clean dependency direction (CLI → orchestrator → pure helpers). The single-source-of-truth ScaffoldEntry[] array is a genuine architectural improvement over the prior parallel paths-plus-writes pattern. However, ArtifactInitType's canonical home in a module named 'init-scaffold' creates ownership ambiguity for a type used cross-cuttingly, and three parallel switch blocks in the same file couple type-enumeration to content-generation in a way that won't scale gracefully with new artifact types.

- **[important]** `src/commands/init-scaffold.ts:1` — **ArtifactInitType defined in scaffold module but used cross-cuttingly across the init subsystem**
  ArtifactInitType is the canonical artifact-type enumeration for the entire init subsystem — it's used in CLI validation (cli.ts), command orchestration (init.ts signature), and scaffold generation (init-scaffold.ts). Its definition now lives in a module whose name implies a narrower scope (scaffold file generation). If the scaffold module is ever refactored or replaced, callers throughout the subsystem would chase a type definition that conceptually belongs at the subsystem boundary, not inside one implementation detail.
  
  Suggestion:

  ```
  Move ArtifactInitType to a shared types module (e.g., src/commands/init-types.ts) that both init-scaffold.ts and init.ts import from, keeping init.ts as the public re-export barrel. This gives the type a neutral home that matches its cross-cutting usage.
  ```

- **[suggestion]** `src/commands/init-scaffold.ts:25` — **Three switch/if-else blocks on ArtifactInitType couple type enumeration to content generation**
  scaffoldEntriesFor (line ~25), buildManifest (line ~57), and buildReadme (line ~140) each contain a switch or if-else chain on ArtifactInitType. Adding a 6th artifact type requires touching three non-adjacent blocks in the same file. A registry map (e.g., Record<ArtifactInitType, { manifest: string, readme: string, entries: ScaffoldEntry[] }>) would consolidate per-type variance into one place and make it immediately obvious when a new type is missing content for one of the builders.
  
  Suggestion:

  ```
  Consider a registry pattern: define a per-type config object containing manifest section, readme template, and extra entries, then have scaffoldEntriesFor, buildManifest, and buildReadme look up that config rather than switching themselves.
  ```

- **[suggestion]** `src/commands/init.ts:33` — **validateTargetDir has non-trivial filesystem logic but is only testable through full init() integration**
  The function handles three distinct error branches (regular file, broken symlink via lstatSync+statSync, permission denied) plus a success path for symlink-to-directory. The PR demonstrates architectural discipline by extracting resolveInitTarget and scaffoldEntriesFor into pure, independently unit-testable modules, but validateTargetDir — with its lstatSync/statSync try/catch pattern — remains a private function tested only indirectly through the full init() command. Its three-branch logic would benefit from the same treatment, especially since the PR added four integration tests specifically exercising its edge cases (file-not-dir, broken-symlink, symlink-to-dir, existing-non-empty).
  
  Suggestion:

  ```
  Extract validateTargetDir into a pure exported function (accepting a targetDir string, returning string | null) in a separate module or as an export from init.ts, then add targeted unit tests for each branch (file, broken symlink, permission error, symlink-to-dir, nonexistent).
  ```

- **[nit]** `src/cli.ts:586` — **Artifact type enumeration duplicated across cli.ts and init-scaffold.ts**
  cli.ts defines `const validTypes = ['skill', 'tool', 'agent', 'prompt', 'pipeline'] as const` for runtime validation, while init-scaffold.ts defines `type ArtifactInitType = 'skill' | 'tool' | 'agent' | 'prompt' | 'pipeline'` as a type alias. These are two non-linked sources of truth. Adding a new type requires updates in two non-adjacent files with no compile-time signal if one location is missed — the type checker won't catch it because the const array isn't derived from the type.
  
  Suggestion:

  ```
  Derive one from the other: define a `VALID_ARTIFACT_TYPES` const array in the shared types module, then derive `type ArtifactInitType = typeof VALID_ARTIFACT_TYPES[number]`. Import the const array in cli.ts for runtime validation and the type everywhere else.
  ```

### Performance
This PR is clean from a performance standpoint. `init()` is a one-shot CLI scaffold command that touches 4–7 files; the sync I/O in `validateTargetDir` (`lstatSync`/`statSync`) and the pre-flight `existsSync` loop block the event loop for microseconds — imperceptible to a human user and not a hot path. No database/network access, no unbounded caches, no closures in render paths, no N+1 patterns. The module split (init-resolve, init-scaffold) adds no runtime overhead since the pure functions are called once per invocation.

_No findings._

### Maintainability
The refactor meaningfully improves maintainability: `resolveInitTarget` is a pure, unit-testable function extracted from CLI wiring, and `scaffoldEntriesFor` consolidates what was previously parallel arrays of paths and inline writes into a single source of truth. The remaining concerns are pre-existing duplication in the content-builder helpers (relocated, not introduced) — specifically `buildReadme`'s identical tool/pipeline cases and near-identical skill/agent/prompt cases.

- **[suggestion]** `src/commands/init-scaffold.ts:221` — **Identical tool/pipeline README templates are copy-paste duplicated**
  The `tool` and `pipeline` cases in `buildReadme` produce verbatim-identical output (~17 lines each). Any future change to the README format (e.g. adding a 'Contributing' section, updating clone paths) must be made in both places, and divergence is likely under time pressure. This duplication was relocated from the old monolithic `init.ts` rather than introduced here, but it now lives in a dedicated content-builder module where the duplication is more visible and has a lower bar for extraction.
  
  Suggestion:

  ```
  Extract a shared `buildInstallableReadme(name, pkg)` helper that both `tool` and `pipeline` (and any future installable artifact types) call.
  ```

- **[suggestion]** `src/commands/init-scaffold.ts:238` — **Near-identical skill/agent/prompt README templates could be parameterized**
  The `skill`, `agent`, and `prompt` cases in `buildReadme` are structurally identical (~22 lines each), varying only in two path segments: the artifact subdirectory (`skill/`, `agent/`, `prompt/`) and the Claude config path (`~/.claude/skills/`, `~/.claude/agents/`, `~/.claude/prompts/`). Copying this block three times means a format change (new section, wording tweak) requires three coordinated edits. Pre-existing but now concentrated in a module where extraction is low-cost.
  
  Suggestion:

  ```
  Replace the three cases with a single `buildIntegratableReadme(name, pkg, artifactDir, claudeDir)` call, mapping each type to its two path parameters.
  ```


---
_Posted by Sage on pi.dev substrate._

PR #154, opened 2026-05-15
feat(login): add --scope option to request packages:write tokens

mellanon:
## Summary

\`arc publish\` requires a \`packages:write\` token, but \`arc login\` only ever requested the marketplace's default scope (\`packages:read\`). The marketplace server has accepted a \`scope\` field on \`POST /api/v1/auth/cli/verify\` since the route was added — but arc never sent it.

Discovered while trying to publish \`@mellanon/code-review\` v0.3.2 to the dev marketplace ([meta-factory#476](https://github.com/the-metafactory/meta-factory/issues/476)). Server-side, \`scope: packages:write\` is supported (verified at \`meta-factory/src/__tests__/cli-auth.test.ts:\\\"supports packages:write scope\\\"\` and \`src/routes/cli-auth.ts:228\`). Client-side, the device-code request body was hardcoded to \`{ device_code: deviceCode }\` with no scope.

Net effect today: **no one can \`arc publish\` via the CLI device-code flow** because read-only tokens reject the publish endpoint, and \`arc login\` unhelpfully reports a stale read-only token as \"Already logged in\".

## What changes

Thread \`scope\` through three layers, all opt-in:

\`\`\`diff
 # src/cli.ts — login command
   .option("-f, --force", "Re-authenticate even if already logged in")
+  .option(
+    "--scope <scope>",
+    "Requested token scope (e.g. packages:read, packages:write). Server defaults to packages:read.",
+  )

 # src/commands/login.ts — LoginOptions
   sourceName?: string;
   force?: boolean;
+  scope?: string;

 # src/lib/device-auth.ts — pollForToken request body
-  body: JSON.stringify({ device_code: deviceCode }),
+  body: JSON.stringify(
+    opts.scope ? { device_code: deviceCode, scope: opts.scope } : { device_code: deviceCode },
+  ),
\`\`\`

When \`--scope\` is not passed, the verify request body is byte-identical to before — so the server's default-to-\`packages:read\` behaviour is preserved and no existing call-site has to change. This is a strict superset of the old flow.

Server-side scope policy stays authoritative. arc just forwards the requested value; the marketplace's \`/auth/cli/verify\` handler validates and rejects anything it doesn't recognise (current allow list: \`packages:read\`, \`packages:write\`).

## Usage

\`\`\`bash
arc login --source metafactory-dev --scope packages:write --force
\`\`\`

After approval through the web flow (where the server confirms the user is authorised for the requested scope), arc receives a write-scoped token, persists it, and \`arc publish\` works against that source.

## Tests

Two new tests in \`test/commands/login.test.ts\` capture the forwarding behaviour:
- \`forwards opts.scope to pollForToken\` — when \`scope\` is provided, it reaches the device-auth layer
- \`omits scope from pollForToken when not provided\` — preserves the legacy request shape

Full suite: 8 pass, 0 fail. \`bunx tsc --noEmit\` clean.

## Why this matters beyond the immediate unblock

- Unblocks [meta-factory#476](https://github.com/the-metafactory/meta-factory/issues/476) — publish code-review v0.3.2 to dev marketplace.
- Prerequisite for [meta-factory#478](https://github.com/the-metafactory/meta-factory/issues/478) — every package migration off REGISTRY.yaml needs \`arc publish\`, which needs a write token.
- Without this, the publishing leg of the marketplace product is non-functional via the CLI auth flow.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -828,16 +828,16 @@ program
   .option("-s, --source <name>", "Target source name (default: first metafactory source)")
   .option("-f, --force", "Re-authenticate even if already logged in")
   .option(
-    "--scope <scope>",
-    "Requested token scope (e.g. packages:read, packages:write). Server defaults to packages:read.",
+    "--token-scope <scope>",
+    "Requested token scope (e.g. packages:read, packages:write). Server defaults to packages:read. Named --token-scope to avoid collision with `arc publish --scope <namespace>`.",
   )
-  .action(async (opts: { source?: string; force?: boolean; scope?: string }) => {
+  .action(async (opts: { source?: string; force?: boolean; tokenScope?: string }) => {
     const paths = createArcPaths();
     const result = await login({
       paths,
       sourceName: opts.source,
       force: opts.force,
-      scope: opts.scope,
+      scope: opts.tokenScope,
     });
 
     if (result.success) {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -827,9 +827,18 @@ program
   .description("Authenticate with metafactory registry (required for publishing only)")
   .option("-s, --source <name>", "Target source name (default: first metafactory source)")
   .option("-f, --force", "Re-authenticate even if already logged in")
-  .action(async (opts: { source?: string; force?: boolean }) => {
+  .option(
+    "--scope <scope>",
+    "Requested token scope (e.g. packages:read, packages:write). Server defaults to packages:read.",
+  )
+  .action(async (opts: { source?: string; force?: boolean; scope?: string }) => {
     const paths = createArcPaths();
-    const result = await login({ paths, sourceName: opts.source, force: opts.force });
+    const result = await login({
+      paths,
+      sourceName: opts.source,
+      force: opts.force,
+      scope: opts.scope,
+    });
 
     if (result.success) {
       console.log(`Logged in to ${result.sourceName}`);
--- a/test/commands/login.test.ts
+++ b/test/commands/login.test.ts
@@ -4,21 +4,49 @@ import { login } from "../../src/commands/login.js";
 import { saveSources } from "../../src/lib/sources.js";
 import type { SourcesConfig } from "../../src/types.js";
 
-// Mock device-auth to prevent real network calls and browser opens
-mock.module("../../src/lib/device-auth.js", () => ({
-  initiateDeviceCode: async () => ({
-    device_code: "test-device-code",
-    user_code: "TEST-CODE",
-    verification_uri: "https://example.com/auth",
-    interval: 1,
-    expires_in: 5,
-  }),
-  pollForToken: async () => ({
-    success: false,
-    error: "Mock: approval timed out",
-  }),
-  openBrowser: () => false,
-}));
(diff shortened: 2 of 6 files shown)


src/cli.ts
```diff
@@ -827,9 +827,18 @@ program
   .description("Authenticate with metafactory registry (required for publishing only)")
   .option("-s, --source <name>", "Target source name (default: first metafactory source)")
   .option("-f, --force", "Re-authenticate even if already logged in")
-  .action(async (opts: { source?: string; force?: boolean }) => {
+  .option(
+    "--scope <scope>",
```
mellanon: **[suggestion/architecture]** Flag name collision with `arc publish --scope`.

`arc publish` already has `--scope <namespace>` (src/cli.ts:893) where the value is a namespace like `@mellanon`. This PR introduces `arc login --scope <scope>` where the value is a token permission like `packages:write`. Same flag name, fundamentally different concepts on adjacent commands in the same CLI surface.

A user who runs `arc publish --scope packages:write` will get a confusing failure (or worse, an unintended publish to a namespace literally named `packages:write`). Conversely `arc login --scope @mellanon` will send `@mellanon` to the server scope allowlist and get rejected.

Options:
- Rename to `--token-scope` on login (most explicit; breaks no existing usage since this flag is new).
- Keep `--scope` but cross-reference both in `--help` text so the distinction is obvious.

Not blocking — the values are disjoint enough that misuse fails loudly — but the ergonomic trap is worth removing before users hit it.

test/commands/login.test.ts
```diff
@@ -107,3 +107,64 @@ describe("login - already logged in", () => {
     expect(result.error).not.toContain("Already logged in");
   });
 });
+
+describe("login - scope option", () => {
+  test("forwards opts.scope to pollForToken", async () => {
+    const captured: { scope?: string } = {};
+
+    // Capture pollForToken args via a per-test mock override
+    mock.module("../../src/lib/device-auth.js", () => ({
```
mellanon: **[suggestion/quality]** `mock.module` cross-test pollution risk.

These two new tests each call `mock.module("../../src/lib/device-auth.js", ...)` inside the `describe` block but never restore the original mock. Bun's `mock.module` is process-global and persists for the rest of the test process. Within this file it's fine because each subsequent call overwrites the previous one and no later test in this file depends on the original baseline behaviour. But it's a foot-gun the next time someone adds a test below these two — the mock baseline at the top of the file (line 8) has been silently replaced.

Fix options:
- Add an `afterEach(() => { /* re-install baseline mock */ })` in the scope-option describe block.
- Or capture the call args via a spy on the existing baseline mock rather than swapping the whole module twice.

Not blocking — works today, all 8 tests pass locally — but the pattern as written is brittle.

mellanon (commented):
Full review (6 lenses + duplication) — fix itself is right; missing manifest version bump blocks ship.

(posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon (commented):
Re-review of fix commit 105a9ac — all three prior findings resolved (1 warning + 2 suggestions), no new findings, 6 lenses + duplication all green.

(posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

PR #167, opened 2026-05-17
fix(symlinks): refuse to overwrite regular files at link target

mellanon:
## Summary

\`createSymlink\` no longer silently unlinks operator-owned regular files at the link target. It now refuses with an actionable error, restoring symmetry with the uninstall path (signal-stack et al. already refuse to delete non-symlinks). Directory backup (\`.pre-arc\` rename) and symlink replacement are unchanged.

## Files

- \`src/lib/symlinks.ts\` — throw on regular file at target
- \`test/unit/symlinks.test.ts\` — new unit tests covering all four cases (symlink replace, directory rename-aside, regular-file refuse, plain create)

## Test plan

- [x] \`bun test test/unit/symlinks.test.ts\` — 7 pass
- [x] \`bun test\` (full suite) — 870 pass, 3 pre-existing skips, 0 fail
- [x] \`bun run tsc --noEmit\` — clean
- [x] Operator data preserved when install would have collided (asserted in test)

Closes #163
Refs #162

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -3,6 +3,7 @@
 import { Command } from "commander";
 import { createArcPaths, ensureDirectories, getDefaultHost } from "./lib/paths.js";
 import { openDatabase } from "./lib/db.js";
+import { SymlinkConflictError } from "./lib/symlinks.js";
 import { install, parseNameVersion } from "./commands/install.js";
 import { list, formatList, formatListJson } from "./commands/list.js";
 import { info, formatInfo, formatInfoJson } from "./commands/info.js";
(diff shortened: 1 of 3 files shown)


mellanon:
Review feedback addressed in 8c54ea9:

- **Typed error**: new `SymlinkConflictError extends Error` with `code: "ARC_SYMLINK_CONFLICT"` and `linkPath: string` field. Callers can now `instanceof`-discriminate.
- **Friendly CLI catch**: top-level `unhandledRejection` listener in cli.ts recognises `SymlinkConflictError` and prints a clean `Error: <message>` + `process.exit(1)`. Other errors keep their existing behaviour. This covers `arc enable`, `arc catalog use`, `wireExtensions`, etc. — paths that previously would have surfaced an unhandled-rejection stack trace.
- **Test tightened**: assertions now check the typed shape (`instanceof SymlinkConflictError`, `code === "ARC_SYMLINK_CONFLICT"`, `linkPath`, message) instead of just regex on the message.

Note: `upgrade-core.ts` already wraps each `createSymlink` call in its own try/catch (per v1 reviewer's trace) and records as a failed step, so no change needed there.

Full suite: 870 pass, 3 skip, 0 fail. Typecheck clean.

PR #188, opened 2026-05-29
fix(upgrade): make registry-extracted packages upgradable (arc#187)

jcfischer:
Fixes #187.

Registry-extracted packages (installed via `@scope/name`, e.g. `@metafactory/soma`) could not be upgraded, breaking the documented SOMA upgrade experience three ways. Root cause for all three: the upgrade path assumed every package is a git clone resolved through the YAML registry index.

## Bug 1 — `arc upgrade <pkg> --check` falsely reports "up to date"
`checkUpgrades` resolved the advertised version via `findInAllSources` (the YAML registry index), but registry packages are published to the metafactory HTTP API. Lookup returned null → fell back to the on-disk manifest version (== installed) → never upgradable.

**Fix:** parse `repo_url`; if it's a package ref, resolve the latest version through `resolveFromRegistry` (metafactory API). Git / YAML-registry packages keep the `findInAllSources` path unchanged.

## Bug 2 — `arc upgrade <pkg> --force` errors `git pull failed: not a git repository`
The pipeline always ran `git pull` in the install dir, but a registry-extracted tarball has no `.git`.

**Fix:** `upgradePackage` now detects registry packages and upgrades them by clean re-download + atomic swap instead of `git pull`. New `fetchAndVerifyRegistryPackage` helper performs the **same verification as install** — SHA-256 + Ed25519 registry signature + Sigstore bundle (reusing the existing verify functions). Git packages keep the `git pull` + pre-pull-HEAD rollback path.

## Bug 3 — remove+install can strand the user
The documented fallback (`arc remove` + `arc install`) can leave the package fully uninstalled when the reinstall fails on a stale token.

**Fix:** bugs 1+2 remove the need for the destructive fallback. Additionally the registry upgrade downloads+verifies **before** touching the working install and restores the prior install on any failure — so an upgrade can never strand the user.

## Verification
- Full suite **958 pass / 3 skip / 0 fail**; `tsc --noEmit` and `eslint` clean.
- New tests: registry `--check` detection, registry `--force` clean re-download (asserts the `git pull` error is gone + DB updated), and a "never strands the user" case (storage 401 mid-upgrade leaves install dir + manifest + DB version intact).

Files: `src/commands/upgrade.ts`, `src/lib/registry-install.ts`, `test/commands/upgrade.test.ts` (+404 / −45).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #189, opened 2026-05-29
fix(ci): green up main — Linux tar bundling + eslint gate

jcfischer:
Two **pre-existing** failures keep `main` CI red (Tests + Lint), independent of any feature branch. Fixing both so feature PRs (e.g. #188) can merge against a green base.

## 1. Tests — `createBundle` fails only on Linux
`tar czf <tarball> … .` wrote the archive **into** `packageDir` while archiving `.`, so the growing tarball perturbed the directory tar was reading. GNU tar (Linux CI) aborts with `file changed as we read it` (exit 1); bsdtar (macOS) tolerates it — which is why the suite passed locally but failed in CI (`createBundle`, `arc bundle`, `arc publish` all route through it).

**Fix:** stage the archive in a temp dir outside the source tree, then `mv` it into place (`mv`, not `fs.rename`, for cross-filesystem moves). Read set is now stable on every platform.

## 2. Lint — `eslint --quiet .` gate has 8 errors
Pre-existing across `commands/jetstream.ts`, `lib/jetstream.ts`, `json-response.ts`, `nats-broker.ts`:
- unnecessary type assertions (`as ConsumerAckPolicy`, `as StreamConfig`, `as ConsumerConfig`)
- unnecessary `?? []` on nats-lib `subjects` (typed non-nullable `string[]`)
- `Array<T>` → `T[]`
- regex `test` → `String#startsWith`
- void-expression arrow needing braces

All mechanical / behavior-preserving. `commands/jetstream.ts` keeps its `ConsumerAckPolicy` alias via annotation rather than assertion (preserves the documented "don't import the runtime enum" intent).

## Verification
- Full suite **955 pass / 3 skip / 0 fail**
- `bunx tsc --noEmit` clean
- `bun run lint:errors` (the CI gate) clean

### DIFF
--- a/src/commands/jetstream.ts
+++ b/src/commands/jetstream.ts
@@ -191,10 +191,11 @@ async function provisionConsumerInternal(
   filterSubject?: string,
 ): Promise<ResourceOutcome> {
   const durable = reviewConsumerName(network, agent);
+  const ackPolicy: ConsumerAckPolicy = "explicit";
   const result = await ensureConsumer(jsm, stream, {
     durable_name: durable,
     ...(filterSubject !== undefined && { filter_subject: filterSubject }),
-    ack_policy: "explicit" as ConsumerAckPolicy,
+    ack_policy: ackPolicy,
   });
   if (!result.ok) {
     throw new ArcNatsCommandError(
--- a/src/lib/bundle.ts
+++ b/src/lib/bundle.ts
@@ -205,13 +205,31 @@ export async function createBundle(
 
   const excludeArgs = buildTarArgs(exclusions, includePatterns);
 
+  // Write the archive to a temp dir OUTSIDE packageDir, then move it into
+  // place. Writing the tarball directly into packageDir while archiving `.`
+  // makes the directory change as tar reads it: GNU tar (Linux CI) aborts
+  // with "file changed as we read it" (exit 1), while bsdtar (macOS) tolerates
+  // it — which is exactly why this passed locally but failed in CI. Staging
+  // outside the source tree keeps the read set stable on every platform.
+  const stageDir = await mkdtemp(join(tmpdir(), "arc-bundle-"));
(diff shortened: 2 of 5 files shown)


jcfischer (commented):
## CodeReview — StandardReview (auto-selected lenses: CodeQuality)

### Lens: CodeQuality

**bundle.ts — tar staging fix**

Root cause correctly diagnosed: `tar czf` writing into its own source directory creates a race on GNU tar. Fix is clean — `mkdtemp` outside source tree, `Bun.spawnSync(["mv", ...])` for cross-fs safety, cleanup on both success and failure paths. Comment explains the *why* well. Imports (`mkdtemp`, `rm`, `tmpdir`) already present at file top.

One observation (not blocking): the `.catch(() => undefined)` on `rm` cleanup is fine for best-effort temp cleanup — consistent with existing pattern at line 244.

**jetstream.ts — annotation over assertion**

`const ackPolicy: ConsumerAckPolicy = "explicit"` preserves the type-narrowing intent without the `as` assertion. Correct approach — ESLint gets a clean pass, and the type annotation still documents the expected shape without the runtime-erasure lie of `as`.

**jetstream.ts, json-response.ts, nats-broker.ts — mechanical lint fixes**

- `?? []` removed on non-nullable `subjects` — correct, nats lib types `string[]`
- `as StreamConfig` / `as ConsumerConfig` removed — underlying types already match
- `Array<T>` → `T[]` — style consistency
- `regex.test()` → `startsWith()` — intent-clearer, marginally faster
- void-expression arrow wrapped in braces — ESLint satisfaction

All behavior-preserving. No semantic changes detected.

### Duplication check

No new code duplicated from existing codebase.

### Summary

Clean CI fix. Both issues (GNU tar race + lint gate) are pre-existing `main` blockers with straightforward, behavior-preserving fixes. 955 tests passing, tsc clean, lint clean. No concerns.

`verdict: blockers=0 majors=0 nits=0 — recommend: approve`

PR #194, opened 2026-06-02
fix: fail fast for unsupported Sigstore publish auth

jcfischer:
Closes #193.\n\nChanges:\n- Reject official Sigstore publishes before upload when the environment would fall back to local cosign device/browser auth.\n- Allow only GitHub Actions OIDC or explicit ARC_SIGSTORE_IDENTITY_TOKEN_FILE / ARC_SIGSTORE_IDENTITY_TOKEN for non-interactive signing.\n- Restrict cosign to the GitHub Actions OIDC provider in CI and keep token values out of process args by writing inline tokens to a temp file.\n\nValidation:\n- rtk bun test test/commands/publish.test.ts test/unit/cosign.test.ts\n- rtk bunx tsc --noEmit\n- rtk bun run lint\n- rtk bun test

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -10,7 +10,7 @@ import {
   ensurePackageExists,
   registerVersion,
 } from "../lib/publish.js";
-import { signSigstoreBundle, type SignSigstoreResult } from "../lib/cosign.js";
+import { resolveSigningAuth, signSigstoreBundle, type SignSigstoreResult } from "../lib/cosign.js";
 import { loadSources, findMetafactorySource } from "../lib/sources.js";
 import { readManifest } from "../lib/manifest.js";
 import type { ArcPaths } from "../types.js";
@@ -104,6 +104,13 @@ export async function publish(opts: PublishOptions): Promise<PublishCommandResul
   let sigstoreBundlePath: string | undefined;
 
   try {
+    if (source.tier === "official" && !opts.allowUnsignedOfficial && !opts.signer) {
+      const signingAuth = resolveSigningAuth();
+      if (signingAuth.kind === "unsupported") {
+        return { success: false, error: `Sigstore signing failed: ${signingAuth.error}` };
+      }
+    }
+
     // 7. Upload tarball
     const uploadResult = await uploadBundle(tarballPath, source, sha256);
     if (!uploadResult.success) {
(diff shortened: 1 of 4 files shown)


jcfischer (commented):
## Review: fix: fail fast for unsupported Sigstore publish auth

**Lenses applied:** CodeQuality, Security, Architecture, Performance

### Summary

Clean, well-scoped fix for #193. The core design — `resolveSigningAuth()` as a pure discriminated-union classifier, fail-fast in both `publish()` (pre-upload) and `signSigstoreBundle()` (pre-cosign-spawn), temp-file for inline tokens with `finally` cleanup — is sound. The fail-closed posture (reject unknown auth rather than falling through to interactive device flow) is the right security call.

### Findings

**Nits (0 blockers, 0 majors, 2 nits):**

1. **Test env save/restore duplication** — The 6-var save/delete/restore pattern is copy-pasted identically across `publish.test.ts` and `cosign.test.ts` (~25 lines each). A small helper like `withCleanSigningEnv(fn)` in `test/helpers/` would collapse both to one-liners and prevent drift if new env vars are added later. Not blocking — the tests are correct as-is.

2. **Missing `resolveSigningAuth` unit tests** — The function is exported and has four return branches, but only the `unsupported` path is tested (indirectly via `signSigstoreBundle`). Direct unit tests for `github-actions`, `identity-token`, and `identity-token-file` branches would catch regressions cheaply — same pattern as the existing `resolveSignerIdentity` tests above. Follow-up material, not blocking.

### What's good

- **Fail-closed by default** — local cosign device/browser auth is explicitly rejected, not silently attempted. Matches the trust model where verification trusts only GitHub Actions OIDC.
- **Token never in process args** — inline `ARC_SIGSTORE_IDENTITY_TOKEN` is written to a 0600 temp file and cleaned up in `finally`. Keeps tokens out of `/proc/*/cmdline`.
- **Pre-upload gate in `publish()`** — catches unsupported auth before the tarball upload, so no wasted registry state on inevitable signing failure.
- **Clean discriminated union** — `SigningAuth` type makes exhaustive matching natural; the `else` branch in `signSigstoreBundle` correctly narrows to `identity-token`.
- **Error message is actionable** — tells the operator exactly what's needed (GH Actions OIDC permissions or explicit token env vars).

**Verdict: approve.** The two nits are follow-up quality — this unblocks `@metafactory/soma` 0.8.1 publishing.

PR #198, opened 2026-06-02
fix: make hook registration idempotent

jcfischer:
## Summary

- make hook registration replace existing matching registrations instead of append-skipping
- reconcile legacy untagged duplicate hook entries with the same matcher and command
- add regression coverage for the duplicate PreToolUse Bash hook shape from #183

Fixes #183.

## Verification

- `rtk bun test test/unit/hooks.test.ts`
- `rtk bun test test/commands/remove-hooks-137.test.ts`
- `rtk bun test test/commands/install.test.ts`
- `rtk bunx tsc --noEmit`
- `rtk bun run lint`
- `rtk bun test`

### DIFF
--- a/src/lib/hooks.ts
+++ b/src/lib/hooks.ts
@@ -28,6 +28,25 @@ interface Settings {
   [key: string]: unknown;
 }
 
+function matcherKey(matcher: string | undefined): string {
+  return matcher ?? "";
+}
+
+function hookGroupHasCommand(entry: SettingsHookGroup, command: string): boolean {
+  return entry.hooks.some((hook) => hook.command === command);
+}
+
+function shouldReplaceHookGroup(
+  packageName: string,
+  hook: { command: string; matcher?: string },
+  entry: SettingsHookGroup,
+): boolean {
+  if (matcherKey(entry.matcher) !== matcherKey(hook.matcher)) return false;
+  if (!hookGroupHasCommand(entry, hook.command)) return false;
+
+  return entry._pai_pkg === packageName || entry._pai_pkg === undefined;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


jcfischer (commented):
## Review: CodeQuality + Architecture

**Lenses applied:** CodeQuality, Architecture

### Analysis

Clean, well-scoped fix for #183 (duplicate hook registrations from `arc install/upgrade`).

**Core change:** `registerHooks` shifts from append-with-skip to filter-then-push (replace-by-tag). Three small helpers (`matcherKey`, `hookGroupHasCommand`, `shouldReplaceHookGroup`) keep the predicate readable and testable.

**Correctness verified:**
- `shouldReplaceHookGroup` matches on `(matcher, command, pkg)` — replaces same-pkg entries AND legacy untagged entries (`_pai_pkg === undefined`), preserves entries from other packages. ✓
- Existing "deduplicates identical hook commands" test still holds: filter removes the old entry, push adds fresh → net 1. Idempotent. ✓
- New test covers the exact scenario from #183: multiple untagged duplicates + one tagged + different matcher + different package → collapsed to 3 entries (OtherPkg/Bash, Cortex/Bash, Read). ✓
- `matcherKey` normalizes `undefined` → `""` via nullish coalescing (also handles `null` from malformed JSON). ✓

**Minor behavior change (acceptable):** Old code was a no-op on re-register (skip if found). New code always removes-then-pushes, rewriting the entry. End state is identical; settings.json gets touched on every install even when unchanged. Not a problem — `arc install` already rewrites settings.json unconditionally.

**No findings.** blockers=0 majors=0 nits=0 — recommend: approve ✅

PR #205, opened 2026-06-04
fix: degrade identityless signed bundles to warn + ARC_SIGSTORE_EXPECTED_IDENTITY escape hatch

### LINKED ISSUE #303: arc 0.30.1 enforces Sigstore but has no expected-signer-identity for the metafactory source — blocks install/upgrade of all signed releases
## Summary

After the cosign/Sigstore signing rollout (soma 0.8.1+), **arc 0.30.1 enforces
Sigstore verification but has no "expected signer identity" to verify against**,
so every signed release fails verification and cannot be installed or upgraded.

Upgrading soma 0.8.0 → 0.8.5:

```
$ arc upgrade soma
Sigstore verification failed: signer_identity missing — cannot verify Sigstore bundle without expected identity
# soma stays at 0.8.0 (non-destructive — good)
```

This is the inverse of the earlier state: 0.8.0 was *"Sigstore signature
MISSING"* (warn + install). Now 0.8.5 IS signed, but arc has nothing to check
the bundle against, and it hard-fails.

## There is no client-side place to supply the identity

```
$ arc source add --help          # only --tier, --type — no signer identity option
$ arc install --help             # only --strict-signing (missing→refuse); no expected-identity flag
$ arc upgrade --help             # no signing flags at all
$ arc source list                # metafactory source carries URL only — no expected signer
$ env | grep -iE "arc|sigstore|signer|cosign|fulcio|identity"   # nothing
```

`arc identity list` principals are **Myelin/NATS** identities, which are a
different thing from the Sigstore/Fulcio signer identity used to sign packages.

## Impact

**Blocks install/upgrade of every signed soma release (0.8.1+) for all users**,
not just one operator. The only installable version is the older unsigned 0.8.0.
Because there's no recovery flag, a user who does `arc remove soma` first (the
documented upgrade workaround) is then **stranded with no soma**, since the
fresh install hits the same verification failure.

## Likely root cause

The cosign signing rollout + arc 0.30.1's stricter verification were deployed
**without the matching expected-signer-identity configuration** in the official
`metafactory` source (or without arc fetching it from the registry).

## Suggested fixes (any one unblocks)

1. Ship/pin the expected Sigstore signer identity in the official `metafactory`
   source config (e.g. the cosign/Fulcio cert-identity + issuer), or have arc
   resolve it from the registry automatically.
2. Provide an `arc install/upgrade --signer-identity <id>` (or source-level
   config) escape hatch.
3. Temporarily downgrade enforcement from hard-fail to warn for signed-but-
   unresolvable bundles until the identity config lands.

## Environment

- arc v0.30.1, installed soma v0.8.0 (active, verified), registry serves v0.8.5
- macOS, cosign present (`/opt/homebrew/bin/cosign`), valid `arc login` token

> Found during an end-to-end release test while upgrading soma to 0.8.5.

jcfischer:
## Context

the-metafactory/soma#303: arc 0.30.1 hard-failed `signer_identity missing — cannot verify Sigstore bundle without expected identity` for **every** signed release, because the registry never populated `signer_identity` (root cause: meta-factory#523, now fixed + backfilled). There was no client-side recovery path — a user who `arc remove`'d first was stranded.

## Change

In `verifyPackageSigstore` when `signature_bundle_key` is present but `signer_identity` is null:

1. **`ARC_SIGSTORE_EXPECTED_IDENTITY` set** → full cosign verification against the operator-supplied identity (verified true/false as normal)
2. **Otherwise** → `verified: null` (degraded) with the escape hatch named in the reason

`verified: null` is the existing warn path: the install command already prints the prominent official-tier warning and `--strict-signing` already turns it into a refusal — so strict installs keep their guarantee while default installs degrade gracefully instead of bricking.

Success reason now names the identity source (`registry` vs `operator-supplied`).

## Tests

- Degraded null + escape hatch named in reason
- Env override → verifier receives operator identity, verified=true
- Env override + cosign mismatch → still verified=false
- Full suite 983 pass / 0 fail, tsc clean

Fixes the-metafactory/soma#303

### DIFF
--- a/src/lib/cosign-verify.ts
+++ b/src/lib/cosign-verify.ts
@@ -27,7 +27,7 @@ export interface VersionSigstoreSigning {
 }
 
 export interface SigstoreVerifyResult {
-  /** true = verified; false = verified-and-rejected; null = not-applicable (unsigned). */
+  /** true = verified; false = verified-and-rejected; null = unverifiable (unsigned, or signed but registry served no identity). */
   verified: boolean | null;
   reason: string;
 }
--- a/src/lib/cosign-verify.ts
+++ b/src/lib/cosign-verify.ts
@@ -93,10 +93,26 @@ export async function verifyPackageSigstore(
     return { verified: null, reason: "version is not sigstore-signed (legacy or unsigned publish)" };
   }
 
-  if (!signing.signer_identity) {
+  // soma#303: a bundle without a registry-served signer_identity is a
+  // registry DATA gap (the bundle exists; only the expected-identity record
+  // is missing), not evidence of tampering. Hard-failing here bricked every
+  // signed install when the registry never populated the field (the
+  // meta-factory#523 extraction bug). Two recovery paths:
+  //   1. ARC_SIGSTORE_EXPECTED_IDENTITY — operator supplies the identity;
+  //      full cosign verification runs against it.
+  //   2. Otherwise degrade to verified=null: callers treat it like the
+  //      unsigned case (prominent warning on official tier, and
+  //      --strict-signing still turns it into a refusal).
+  const operatorIdentity = process.env.ARC_SIGSTORE_EXPECTED_IDENTITY;
+  const expectedIdentity = signing.signer_identity ?? operatorIdentity ?? null;
+  const identitySource = signing.signer_identity ? "registry" : "operator-supplied";
+
+  if (!expectedIdentity) {
     return {
-      verified: false,
-      reason: "signer_identity missing — cannot verify Sigstore bundle without expected identity",
+      verified: null,
+      reason:
+        "bundle present but registry served no signer_identity — cannot verify; " +
+        "set ARC_SIGSTORE_EXPECTED_IDENTITY=<workflow identity> to verify against a known identity",
     };
   }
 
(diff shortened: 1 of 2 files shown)


jcfischer (commented):
## Code Review — arc#205

**Lenses applied:** CodeQuality, Security, Architecture

### Summary

Clean, well-scoped fix. The `verified: false` → `verified: null` change correctly maps the identityless-bundle case to the existing degraded/warn path that callers already handle. Two recovery paths (env override for operators, graceful degrade for everyone else) are properly layered. `--strict-signing` continues to refuse on null, so the strict guarantee is preserved.

### Findings

**Nit (1):** The `SigstoreVerifyResult` docstring at `src/lib/cosign-verify.ts:30` says `null = not-applicable (unsigned)` — but the new degraded path returns null for a *signed* bundle missing its identity, not just unsigned versions. The semantics now include "signed-but-unverifiable" in the null bucket. Consider updating the docstring to something like:

```
/** true = verified; false = verified-and-rejected; null = unverifiable (unsigned or missing identity). */
```

This is cosmetic — callers already behave correctly.

### What's good

- **Security model intact.** Confirmed both callers (`cli.ts:284`, `registry-install.ts:587`) only fail-closed on `verified === false`. The null→warn→`--strict-signing` chain works as claimed.
- **Env var naming.** `ARC_SIGSTORE_EXPECTED_IDENTITY` (install-side) is clearly distinct from `ARC_SIGSTORE_SIGNER_IDENTITY` (publish-side, `cosign.ts:167`). No collision risk.
- **Test coverage.** Three new tests cover the exact decision branches: degraded null, env-override success, env-override mismatch. Env cleanup in `finally` blocks prevents leakage.
- **Comments.** The block comment at lines 96–105 explains the *why* (registry data gap, not tampering) thoroughly — anyone reading this code later will understand the design choice.
- **Reason strings.** Both the degraded reason and the success reason name the escape hatch / identity source, making debug output actionable.

blockers=0 majors=0 nits=1 — recommend: approve

PR #206, opened 2026-06-04
fix(#204): 60s register timeout + poll submission to terminal status on abort

jcfischer:
## Root cause

`POST /versions` validates in-band since meta-factory#300 (~9s for soma's 8 MB / 275 files). The shared `API_TIMEOUT_MS = 10_000` aborted right on that edge → client disconnect → Cloudflare cancels the request context → submission stranded at `validating` → reconciler system-rejects. Soma 0.8.2/0.8.3/0.8.4 each match exactly one `clientDisconnected` in Workers analytics. Worse: the retry run's 409-probe reported the doomed `validating` submission as **success** — green CI, package never landed.

## Changes

1. **`REGISTER_TIMEOUT_MS = 60_000`** dedicated to the registration POST (rationale comment explains the in-band validation budget)
2. **Abort honesty:** on `TimeoutError`/`AbortError`, poll the per-version submission endpoint (5s cadence / 90s budget, injectable) —
   - terminal status reached (`pending_review`/`approved`/`rejected`) → report THAT as the outcome
   - still non-terminal at deadline → `success: false`, names submission id + status + poll URL
   - no submission visible → `success: false`, advises retry (request never reached registry)
3. Non-timeout network errors unchanged

409-probe path intentionally untouched: with the reconciler live, a cross-run probe sees a terminal status within ~5 min, and same-run timeouts are now handled by the poll above.

## Tests

4 new: timeout→poll→terminal success; timeout→stuck-validating honest failure; timeout→no-submission retry guidance; non-timeout errors keep plain reporting. Full suite 987 pass / 0 fail, tsc clean.

Fixes #204

### DIFF
--- a/src/lib/publish.ts
+++ b/src/lib/publish.ts
@@ -15,6 +15,20 @@ import { README_VARIANTS } from "./bundle.js";
 const API_TIMEOUT_MS = 10_000;
 const UPLOAD_TIMEOUT_MS = 120_000;
 
+// arc#204: POST /versions runs L1+L2 validation IN-BAND on the registry
+// (meta-factory#300) — ~9s for an 8 MB package, more under load. The shared
+// 10s API timeout sat right on that edge: aborting mid-request disconnects
+// the client, Cloudflare cancels the request context, and the submission is
+// stranded at `validating` (then system-rejected by the reconciler). Give
+// registration its own budget well clear of the in-band validation time.
+const REGISTER_TIMEOUT_MS = 60_000;
+
+/** Submission statuses that end the publish-side wait (human review is next, or it's decided). */
+const TERMINAL_SUBMISSION_STATUSES = new Set(["pending_review", "approved", "rejected"]);
+
+const SUBMISSION_POLL_INTERVAL_MS = 5_000;
+const SUBMISSION_POLL_TIMEOUT_MS = 90_000;
+
 // ── Debug logging ────────────────────────────────────────────
 
 function debugLog(msg: string): void {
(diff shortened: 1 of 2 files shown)


jcfischer (commented):
**SOP: pr-review | PR: the-metafactory/arc#206 | Lenses: CodeQuality, Security, Architecture**

## Verdict: APPROVE (comment — can't self-approve)

Clean, well-scoped fix for a real production bug (soma 0.8.2–0.8.4 silent publish failures). The root cause analysis in the PR body is excellent — the chain from client timeout → CF request context cancellation → stranded `validating` submission → retry 409-probe false-positive is clearly traced.

### CodeQuality ✅

- **TERMINAL_SUBMISSION_STATUSES** set is well-chosen (`pending_review`, `approved`, `rejected`). Notably `validating` and `audit` are correctly excluded — they're intermediate.
- `pollSubmissionToTerminal` loop structure is clean: deadline-based with injectable intervals for testing. The `for(;;)` with a deadline check is idiomatic for polling.
- Three-way return from timeout handler (terminal → success, non-terminal → honest failure with status, no submission → retry guidance) covers all observable states completely.
- Test coverage hits all three branches plus the non-timeout regression guard. `pollIntervalMs: 1` / `pollTimeoutMs: 20` injection for tests is clean — no sleepy CI.
- `opts` parameter is optional with defaults — backward compatible. Single caller in `publish.ts:155` doesn't pass it, no call-site change needed.

### Security ✅

- No new auth, input handling, or external data trust changes. Poll uses same `buildPublishHeaders(source)` as original POST — no credential leak or downgrade.
- `AbortSignal.timeout(API_TIMEOUT_MS)` on poll GETs prevents the polling loop itself from hanging on a single request.

### Architecture ✅

- Separation of concerns: `pollSubmissionToTerminal` is a private helper, `RegisterVersionOpts` is the public surface. Poll function doesn't touch `registerVersion`'s response shaping — just observes server state.
- The `statusCode: 200` on the recovered-terminal path is a pragmatic lie (the original POST never returned 200), but callers only branch on `success` — harmless in practice.

### Nits (non-blocking)

None. Tight bug fix with clear provenance.

**blockers=0 majors=0 nits=0 — recommend: approve.**

PR #220, opened 2026-06-09
fix(install): #219 — cross-platform path guard so Windows installs don't false-reject

mellanon:
## Summary

Fixes #219 — two separator-naive bugs in the `arc install` path guard (`src/commands/install.ts`) that broke **every** `arc install` on Windows. macOS/Linux were unaffected, which is why it went unnoticed.

| Line | Bug | Effect on Windows |
|------|-----|-------------------|
| L113 | `opts.preExtractedPath.split("/").pop()` — `split("/")` doesn't split on `\` | `.pop()` returned the **whole** path, so the registry error echoed the full path as the "repo name" |
| L120 | `normalizedInstall.startsWith(normalizedRepos + "/")` — hardcoded `/` | `join()` returns `\`-separated paths on Windows, so a valid child (`C:\…\repos\soma`) never matched the `/`-suffixed prefix → rejected with `Refusing to install: repo name "…" would escape repos directory` |

## The fix

Refactor the containment + repo-name logic out of the install path into two **pure helpers** in `src/lib/repo-name.ts`, each accepting a *path flavor* (a `Pick` of the Node `path` API):

- `isInsideRepos(reposDir, installPath, p = path)` — replaces the `startsWith` prefix check with a `path.relative`-based containment check. `inside` ⇔ `rel === "" || (!rel.startsWith("..") && !p.isAbsolute(rel))`. This keeps the `..`-escape protection cross-platform while accepting valid nested paths, and additionally blocks cross-drive Windows paths (relative across roots is absolute).
- `repoNameFromPreExtracted(preExtractedPath, p = path)` — replaces `split("/").pop()` with `p.basename(...)` (separator-safe on both flavors), trims a trailing separator, and returns `undefined` so the caller falls back to `extractRepoName(repoUrl)`.

The install call sites pass the **platform default** `path`, so posix behavior is byte-for-byte identical (the full 662-test unit suite is green, including 4 new posix guard cases — no regression).

## Why this is verifiable without a Windows machine

The operator **cannot test on Windows**, so the cross-platform unit tests ARE the verification. On a macOS/Linux host the platform-default `path` resolves to posix, so a naive test can't prove the Windows behavior. By threading a `PathFlavor` through the pure helpers, the new test file (`test/unit/install-path-guard.test.ts`) calls them with **both `path.win32` and `path.posix`**, genuinely exercising the `\`-separator path on this macOS host.

The win32 assertions (all passing on macOS CI):

1. **win32** valid child `C:\…\repos\soma` → inside (the #219 git-URL repro now passes)
2. **win32** scoped child `C:\…\repos\metafactory__soma` → inside (the #219 registry repro)
3. **win32** `..`-escape `C:\…\repos\..\evil` → NOT inside (escape protection intact on Windows)
4. **win32** sibling-prefix `…\repos-evil\soma` → NOT inside (no false-prefix match)
5. **win32** repo-name `C:\…\repos\metafactory__soma` → `"metafactory__soma"` (not the whole path — fixes the registry-error repro)
6. **posix** valid nested → inside; `..`-escape → NOT inside; repo-name `/…/repos/soma` → `"soma"` (no regression)
7. Boundary: install path === repos dir → inside

## Acceptance criteria (from #219)

- [x] `arc install @metafactory/soma` resolves under reposDir → install proceeds on Windows (registry repro now passes the guard)
- [x] `arc install <git-url>` resolves under reposDir → install proceeds on Windows
- [x] `..`-escape attempts still blocked on **all** platforms
- [x] Unit tests added for Windows path separators in install path validation

## Verification

- `bunx tsc --noEmit` → 0 errors
- `bunx eslint` on touched files → clean
- `bun test test/unit/` → 662 pass, 0 fail (includes the 13 new cross-platform guard cases)

### DIFF
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -27,7 +27,7 @@ import {
   rollbackArtifactSymlinks,
 } from "../lib/artifact-installer.js";
 import { wireExtensions } from "../lib/extensions.js";
-import { extractRepoName } from "../lib/repo-name.js";
+import { extractRepoName, isInsideRepos, repoNameFromPreExtracted } from "../lib/repo-name.js";
 import { requireBrokerForManifest } from "../lib/nats-broker.js";
 import {
   type HostOverrides,
@@ -108,16 +108,17 @@ export interface InstallResult {
 export async function install(opts: InstallOptions): Promise<InstallResult> {
   const { arc, host, db, repoUrl } = opts;
 
-  // 1. Clone repo (or use pre-extracted path for registry installs)
-  const repoName = opts.preExtractedPath
-    ? opts.preExtractedPath.split("/").pop() ?? extractRepoName(repoUrl)
-    : extractRepoName(repoUrl);
+  // 1. Clone repo (or use pre-extracted path for registry installs).
+  // basename (via repoNameFromPreExtracted) is separator-safe — `split("/")`
+  // returned the whole path on Windows `\`-separated paths (#219).
+  const repoName = repoNameFromPreExtracted(opts.preExtractedPath) ?? extractRepoName(repoUrl);
   const installPath = opts.preExtractedPath ?? join(arc.reposDir, repoName);
 
-  // S2: Path traversal guard — ensure installPath stays inside reposDir
-  const normalizedInstall = join(installPath); // resolves ../ segments
-  const normalizedRepos = join(arc.reposDir);
-  if (!normalizedInstall.startsWith(normalizedRepos + "/") && normalizedInstall !== normalizedRepos) {
+  // S2: Path traversal guard — ensure installPath stays inside reposDir.
+  // Uses a path.relative-based containment check (isInsideRepos) instead of a
+  // separator-naive `startsWith(reposDir + "/")` that false-rejected valid
+  // `\`-separated Windows child paths while still blocking `..` escapes (#219).
+  if (!isInsideRepos(arc.reposDir, installPath)) {
     return {
       success: false,
       error: `Refusing to install: repo name "${repoName}" would escape repos directory`,
--- a/src/lib/repo-name.ts
+++ b/src/lib/repo-name.ts
@@ -1,3 +1,64 @@
+import nodePath from "path";
+
+/**
+ * A subset of the Node `path` API selectable per OS flavor.
+ *
+ * The install path guard must behave correctly on the OS it runs on, but the
+ * platform-default `path` resolves to posix on a macOS/Linux CI — so the
+ * Windows (`\`-separator) behavior would be untestable there. By threading a
+ * `PathFlavor` through the pure helpers below, the install call sites use the
+ * platform default (`nodePath`) while tests can pass `path.win32` / `path.posix`
+ * explicitly to prove cross-platform correctness on any host. See issue #219.
+ */
+export type PathFlavor = Pick<typeof nodePath, "basename" | "relative" | "isAbsolute">;
+
+/**
+ * Containment guard: is `installPath` the repos dir itself, or a path nested
+ * inside it? Uses `path.relative` so a `..`-escape (the relative path starts
+ * with `..`) or an absolute relative path (different drive/root on Windows) is
+ * rejected, while a legitimately nested child is accepted — on every separator
+ * flavor. Replaces the separator-naive `startsWith(reposDir + "/")` check that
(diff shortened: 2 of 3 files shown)


mellanon (commented):
## Independent code review — PR #220 (fix #219 cross-platform install path guard)

**Verdict: LGTM — would-approve.** (Posted as a comment because the `gh` token identity is the same as the PR author, so `--approve` is blocked as self-approval; this is an independent second-pass review, not the author's self-review.)

Reviewed fresh and skeptically at PR head (`6cfaf1f`) with the CodeReview skill — CodeQuality + Security + Architecture lenses. I read the changed files, ran the tests, type-checked, eslinted, and independently probed the containment logic (including cases the author did **not** test) plus simulated the **old** code against the win32 inputs to confirm the new tests are real regression guards.

### Correctness of the containment logic — verified complete

`isInsideRepos`: `rel === "" || (!rel.startsWith("..") && !p.isAbsolute(rel))`. I probed both flavors beyond the committed test cases. All behave correctly:

| Case (win32 + posix) | Result | Notes |
|---|---|---|
| valid child / deep child | INSIDE | correct |
| repos dir itself (boundary) | INSIDE | `rel === ""` |
| `..`-escape (direct + buried `/a/../../evil`) | rejected | `relative` resolves `..`, then `startsWith("..")` catches it |
| child-then-back-in (`/a/../soma`) | INSIDE | correct — normalization-aware |
| sibling-prefix `repos-evil/soma` | rejected | the **old** `startsWith(reposDir + "/")` had this bug; new code fixes it |
| exact-prefix-no-sep `repos2` | rejected | also a latent old-code class; correct now |
| cross-drive `D:\…\repos\soma` (win32) | rejected | `relative` across roots is absolute → `isAbsolute` catches it |
| mixed separator `…\repos/soma` (win32) | INSIDE | win32 treats both `/` and `\` as separators |
| trailing separator | INSIDE / correct name | handled |

No escape bypass found on either flavor. The `..`-resolution that the old code did explicitly via `join(installPath)` is now handled implicitly by `path.relative` — my buried-`..` probes confirm this is sound.

`repoNameFromPreExtracted`: `basename` after trimming trailing separators; returns `undefined` for empty/root-only inputs (`C:\`, `/`) so the caller falls back to `extractRepoName(repoUrl)`. Correct.

### win32 tests are genuine regression guards (not tautologies)

I simulated the **old** code against the committed win32 inputs:
- old containment (`join` + hardcoded `+ "/"`): valid win32 child → `false` (false-reject). New asserts `true`. → test fails old code. ✅
- old repo-name (`split("/").pop()`): win32 path → returns the **whole path**. New asserts `"metafactory__soma"`. → test fails old code. ✅

Both new test groups use `path.win32` / `path.posix` explicitly, so they exercise the `\`-separator behavior independent of the macOS host. Real guards.

### No posix regression

Production call sites (`install.ts:114,121`) pass the platform-default `path`. `reposDir` is always absolute (`paths.ts:99` → `join(configRoot, "pkg", "repos")`) and `preExtractedPath` comes from the registry extract pipeline (absolute), so the `path.relative`-resolves-against-cwd edge for relative inputs is **not reachable** in production. Full suite: **662 pass / 0 fail**. macOS/Linux behavior byte-for-byte unchanged.

### Verification run locally (this review, not trusting the PR body)

- `bun test test/unit/install-path-guard.test.ts` → 13 pass / 0 fail
- `bun test test/unit/` → 662 pass / 0 fail
- `bunx tsc --noEmit` → 0 errors
- `bunx eslint` on the three touched files → clean
- Independent edge-case probe (17 cases, both flavors) → all correct
- Old-code simulation against win32 inputs → confirms tests are real guards

### Findings

- **Blockers:** none.
- **Majors:** none.
- **Minor (non-blocking, optional):** the refactor drops the explicit `join(installPath)` pre-normalization the old code had; correctness now relies on `path.relative` resolving `..` internally. It does (verified), but a one-line note in the `isInsideRepos` doc that it depends on `relative` performing the normalization would make the contract explicit for future readers. Not worth holding the PR.

**Merge-readiness: ready to merge.** Fix is minimal, correctly scoped to #219, escape protection is provably intact on both separator flavors, and the cross-platform tests are legitimate regression guards. All four acceptance criteria from #219 are met.

— Independent review by Marcus (Engineer agent). Approval intent recorded here as a comment due to the self-approval block; a non-author maintainer should click Approve to satisfy the ruleset.

PR #230, opened 2026-06-10
feat(manifest): add type: process arc-manifest schema (dev-loop F-6d)

mellanon:
## What

dev-loop **F-6d** (arc side) — the minimal `type: process` type-enum + manifest validation. Promotes DD-47's process-node taxonomy (Deterministic [D] / Agentic [A] / Human gate [H]) to a first-class arc-manifest, validated at parse time.

Schema spec (the normative authority for this PR): **the-metafactory/meta-factory#551** (`design/dev-loop-process-schema.md`).

## Changes

- **`src/types.ts`** — add `process` to `ArtifactType` and the `ArcManifest.type` union; add `ProcessSpec`, `ProcessStep`, `ProcessAgentStep`, `ProcessGateStep`, `ProcessAgentSovereignty`, `ProcessAgentFederation`, mirroring pulse `src/types/pipeline.ts` + `src/types/agent.ts`.
- **`src/lib/manifest.ts`** — add `validateProcess`; make `capabilities` optional for `type: process` (a process declares deps per-node).
- **`test/unit/manifest-process.test.ts`** — 11 tests (RED→GREEN), incl. round-tripping pulse's real cross-principal `P_BUILD_JOURNAL`.

## The schema DESCRIBES pulse reality, not the issue's idealised DAG

The framing contract (meta-factory#550) requires the schema to describe what **pulse actually executes**. A pulse process is an **ordered `actions:` array**: a bare string is a [D] action ref, an `agent:` map is an [A] step, a `gate:` map is an [H] step. This is **not** the explicit-DAG (`nodes`/`startNode`/`endNodes`/`dependsOn`, ISO-8601 timeouts, gate `approvalRoles`, agent `costCap`/`maxRetries`) sketched in the issue body — that shape does not round-trip a real pulse pipeline. Full divergence table + rationale: meta-factory#551 §4.

Concretely, two consequences for this PR:
- **No cycle check.** There are no edges in an ordered array, so the issue's `hasCycle`/topological-sort logic has nothing to operate on. (A branching-DAG future is tracked — meta-factory#551 §6.)
- **`timeout_ms` is validated as a positive number of milliseconds**, not an ISO-8601 duration — pulse uses ms throughout; validating `PT1H` would reject every real pulse definition.

## Validation rules (`validateProcess`)

1. `process` block present, with a non-empty `name` and a **non-empty** `actions` array.
2. Every step is a bare non-empty string or a **single-keyword** map.
3. Keyword ∈ {`agent`, `gate`, `map`, `filter`, `reduce`, `parallel`, `retry`}.
4. `agent:` steps need `name` + `capability` + `prompt`.
5. `gate:` steps need `name` + `prompt`.
6. `timeout_ms`, when present, is a positive number (ms).

Composition primitives (`map`/`filter`/…) pass opaquely — pulse owns their runner semantics (Anti-Abstraction Gate).

## Round-trip (one of the 11 tests)

`P_BUILD_JOURNAL` (pulse `examples/build-journal/pipeline.yaml`) validates unchanged: 8 ordered steps — 5 [D] action refs, 2 [A] agent steps (one federated to `andreas`/`meta-factory`), 1 [H] gate — `sovereignty` + `federate` preserved exactly.

## Scope guard

Does **NOT** touch `src/commands/install.ts`. Capability-resolution at install time (the issue's "agent step with missing capability fails install") crosses into install and is owned by another lane — noted as a follow-up in meta-factory#551 §6, not implemented here.

## Gates

- `bunx tsc --noEmit` ✅
- `bun run lint` (eslint) ✅
- `bun test test/unit/` ✅ (686 pass, 0 fail; the new file adds 11)

Refs cortex#835, cortex#869, meta-factory#550

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -3,7 +3,7 @@
 // ── Catalog types ──────────────────────────────────────────────
 
 /** Artifact types in the catalog */
-export type ArtifactType = "skill" | "agent" | "prompt" | "tool" | "component" | "pipeline" | "rules" | "library" | "action";
+export type ArtifactType = "skill" | "agent" | "prompt" | "tool" | "component" | "pipeline" | "process" | "rules" | "library" | "action";
 
 /** Catalog entry type — controls trust level and install behavior */
 export type CatalogEntryType = "builtin" | "community" | "system" | "custom";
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -3,7 +3,7 @@
 // ── Catalog types ──────────────────────────────────────────────
```
mellanon, line 3: Pre-existing observation surfaced by this PR: `ArtifactType` union (line 4 of types.ts) and the `ArcManifest.type` union (now line 363) are diverged — `ArtifactType` includes `"action"` but not `"system"`; `ArcManifest.type` includes `"system"` but not `"action"`. This PR correctly adds `"process"` to BOTH, which is the right pattern. The pre-existing divergence isn't introduced here and is a separate cleanup, but noting it for awareness.

mellanon (commented):
Reviewed as part of the F-6d two-PR logical change (meta-factory#551 + arc#230).

### Divergence correctness: ordered-actions is right

I read pulse's `src/types/pipeline.ts` directly. `PipelineStep` is a discriminated union of bare string | `{ map }` | `{ filter }` | `{ reduce }` | `{ parallel }` | `{ retry }` | `{ gate }` | `{ agent }` — exactly the shape this PR validates. No `id`, no `dependsOn`, no `startNode`/`endNodes`. The issue's explicit-DAG sketch would have required pulse pipelines to be rewritten to pass validation. Dropping `hasCycle`/topological-sort is correct for v1 — an ordered array has no edges to cycle-check.

### Validation fidelity: tight

`validateProcess` runs `unknown`-first validation throughout rather than trusting the optimistic YAML-parse cast. All edge cases I checked are caught:

- `process: null` → `isRecord(null) = false` → caught
- `process: []` → `isRecord([]) = false` → caught
- multi-key step maps → `keys.length !== 1` → caught
- empty string D-step → `step.length === 0` → caught
- `timeout_ms: 0` → `0 <= 0` → caught
- `timeout_ms: "PT1H"` → `typeof !== "number"` → caught
- unknown keyword (`frobnicate:`) → `KNOWN_PROCESS_STEP_KEYWORDS.has` → caught

**One validation gap (nit, not blocking):** The composition primitives (`map`, `filter`, `reduce`, `parallel`, `retry`) are accepted opaquely. A `map:` step containing a nested `actions:` array with invalid agent/gate steps will pass arc validation. This is **intentional** per the Anti-Abstraction Gate (pulse owns runner semantics for its composition primitives) and the comment in the code says so. Noting it here so the gap is visible in the review record rather than discovered later.

### Round-trip proof: structurally complete

`P_BUILD_JOURNAL` test uses inline YAML fixture, not a live read of `pulse/examples/build-journal/pipeline.yaml`. The fixture is structurally faithful — 8 steps, 5D+2A+1H, `federate`+`sovereignty` on the perspective step. Assertion goes beyond `.toHaveLength(8)`: it explicitly counts strings (D), `agent` maps (A), and `gate` maps (H) by type-check. That's the right level of proof for a round-trip test.

The gap (can't catch pulse YAML drift at test time) is a cross-repo dependency limitation, not a test quality failure.

### Scope guard: clean

3 files changed: `src/types.ts`, `src/lib/manifest.ts`, `test/unit/manifest-process.test.ts`. `src/commands/install.ts` not touched. Capability-resolution at install is correctly deferred.

### Pre-existing observation (inline comment on types.ts)

`ArtifactType` and `ArcManifest.type` have pre-existing members that diverge from each other (`"action"` in ArtifactType only; `"system"` in ArcManifest.type only). This PR correctly adds `"process"` to both. The pre-existing divergence is out of scope here — separate cleanup.

### Gates: all green

3/3 checks pass (typecheck + lint + unit tests with 686 pass, 0 fail + 11 new).

---

**blockers=0 majors=0 nits=2**

- nit/types.ts: Pre-existing `ArtifactType` vs `ArcManifest.type` divergence (separate cleanup, not this PR).
- nit/round-trip: Opaque composition-primitive validation (intentional, documented in code, just worth noting in the review record).

Neither blocks merge. The implementation is correct, tight in scope, and well-tested.

blockers=0 majors=0 nits=2 — recommend: approve (cannot self-approve; submit from a second reviewer account)

PR #231, opened 2026-06-10
feat(install): F-6c — ordered library installs + atomic rollback

mellanon:
## What

dev-loop **F-6c**: `type: library` installs now (1) land artifacts in **dependency order**, (2) **roll back atomically in reverse order** when an artifact fails mid-sequence, (3) report **partial per-artifact state** via a transactional journal, and (4) support **resume** from a named artifact. Lifts the arc#140 P4 single-package rollback model to multi-artifact library sequences.

Closes #227
Refs cortex#835 (agentic dev pipeline — design §6.4 F-6c)

## Design as built

**1. Dependency ordering** — `toposortArtifacts()` in `src/lib/artifact-installer.ts`. Kahn's algorithm over `depends_on.packages` edges, but **only intra-library edges constrain order** — a `depends_on.packages` entry naming a package not contained in this library (an external tool/repo dep) is an install-time gate handled elsewhere, not an ordering edge, so it is ignored. Independent peers keep declaration order (deterministic journal/console). A cycle (direct, self, transitive) throws with the offending names before any filesystem mutation.

**2. Transaction journal** — `beginLibraryInstallTransaction()` in `src/lib/install-transaction.ts`. Rather than re-deriving rollback from an evidence blob, the library transaction **holds each successful artifact's live `InstallTransaction` handle** (the exact arc#140 P4 sub-transaction) plus a `removeDbRow` callback. `rollback()` walks the landed artifacts in **reverse order**, calling each sub-transaction's `rollback()` (symlinks/hooks/launchd/extensions) and then `removeDbRow` for the committed DB row the sub-transaction itself never had to undo (the DB commit is the last step of a single-artifact install). Skipped (pre-existing) artifacts are **never** rolled back. New types: `ArtifactInstallState` (skipped/success/failed/rolled_back), `ArtifactInstallDetail`, `LibraryInstallJournal`.

**3. Library install logic** — `installLibrary()` in `src/commands/install.ts`: toposort → ordered install → **stop on first failure** → reverse-order rollback → journal. `installSingleArtifact()` gains an optional `onTransaction` hook so the library can capture the live sub-transaction; **standalone installs are unaffected** (callback absent). `InstallResult` gains `journal` (authoritative per-artifact state) while keeping `artifacts: InstallResult[]` for backward compatibility.

**4. Resume** — `InstallOptions.resumeFromArtifact`: skips every artifact ordered before the named one (assumed already installed/skipped) and installs from it onward with the same ordered + atomic-rollback semantics. Unknown name → clear error.

## Divergences from spec (followed the code's reality)

1. **`InstallResult.artifacts` kept as `InstallResult[]`** (spec proposed switching it to `ArtifactInstallDetail[]`). Existing library tests read `.success`/`.name`/`.version` off it; changing the shape would break acceptance criterion #6 (backward compat). The spec's intent — authoritative per-artifact state — is delivered via the **new `journal` field** (`journal.artifacts: ArtifactInstallDetail[]`).
2. **Journal/state types live in `lib/install-transaction.ts`, re-exported from `types.ts`** (spec placed `ArtifactInstallState`/`ArtifactInstallDetail` in `types.ts`). `types.ts` imports nothing from `lib/`, and the journal references `InstallTransactionEvidence` (a lib type); defining there + re-exporting keeps the lib→types dependency arrow one-directional and avoids an import cycle. They are still importable from `../types.js`.
3. **Rollback model holds live sub-transaction handles, not an evidence blob** (spec sketched an evidence-replay model). Reusing the live `InstallTransaction.rollback()` is the literal "extend, don't fork" of arc#140 P4 — one rollback path, not two.
4. **`PackageDependency` is `{ name, repo }`** (no `version` field) in this repo; `depends_on.packages[].name` is the ordering key, matching the spec's `.map(p => p.name)`.

## Tests (TDD — 19 new)

- `test/unit/artifact-toposort.test.ts` (8): no-deps passthrough, dependency-before-dependent, transitive chain, diamond, external-dep ignored, direct cycle, self-cycle, deterministic peer order.
- `test/unit/library-install-transaction.test.ts` (4): journal pending state, skip/success/failure recording, reverse-order rollback + DB-row removal, skipped-not-rolled-back.
- `test/commands/library-install-ordering.test.ts` (7): dependency-ordered install, reverse-order rollback (DB **and** filesystem symlinks), stop-on-first-failure, resume-from-artifact, resume-unknown-name rejection, cycle reporting (no filesystem touch), already-installed skip.

Test helper `createMockLibraryRepo` extended with per-artifact `dependsOn` + `postinstall` (to inject mid-sequence failures).

## Gates

```
$ bun run lint:errors
$ eslint --quiet .
# exit 0 — clean

$ bunx tsc --noEmit
# exit 0 — clean

$ bun test
 1048 pass
 3 skip
 0 fail
 2733 expect() calls
Ran 1051 tests across 69 files.
```

(The 3 skips are pre-existing and unrelated; 0 fail. CI's `test` job is `continue-on-error` per arc#151 — lint + typecheck are the gating jobs and both pass.)

### DIFF
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -12,7 +12,7 @@ import type {
 import type { Database } from "bun:sqlite";
 import { errorMessage } from "../lib/errors.js";
 import { readManifest, readLibraryArtifacts, assessRisk, formatCapabilities } from "../lib/manifest.js";
-import { recordInstall, getSkill } from "../lib/db.js";
+import { recordInstall, getSkill, removeSkill } from "../lib/db.js";
 import { runScript, runLifecycleScripts } from "../lib/scripts.js";
 import {
   registerHooks,
@@ -25,6 +25,7 @@ import {
   resolveArtifactSourceDir,
   installNodeDependencies,
   rollbackArtifactSymlinks,
+  toposortArtifacts,
 } from "../lib/artifact-installer.js";
 import { wireExtensions } from "../lib/extensions.js";
 import { extractRepoName, isInsideRepos, repoNameFromPreExtracted } from "../lib/repo-name.js";
@@ -41,8 +42,12 @@ import {
 } from "../lib/hosts/launchd-install.js";
 import { isDarwinLaunchdHost } from "../lib/hosts/darwin-launchd.js";
 import {
+  ArtifactInstallState,
   beginInstallTransaction,
+  beginLibraryInstallTransaction,
+  type InstallTransaction,
   type InstallTransactionEvidence,
+  type LibraryInstallJournal,
 } from "../lib/install-transaction.js";
 
 export interface InstallOptions {
@@ -64,6 +69,16 @@ export interface InstallOptions {
   artifactName?: string;
   /** When installing from a library, the library name (for DB tracking) */
   libraryName?: string;
+  /**
+   * Resume a failed library install from the named artifact (arc#227 / F-6c).
+   *
+   * Skips every artifact that orders BEFORE this one in the dependency-sorted
+   * sequence (they are assumed already installed or deliberately skipped), then
+   * installs from this artifact onward with the same ordered / atomic-rollback
+   * semantics. Library installs only; ignored for standalone installs.
+   * Example: `arc install dev-loop --resume-from=dev` after fixing the broker.
+   */
+  resumeFromArtifact?: string;
   /**
    * Pre-extracted install path (for registry installs from F-4).
    * When provided, skips git clone and uses this directory as the source.
@@ -89,8 +104,20 @@ export interface InstallResult {
   error?: string;
   manifest?: ArcManifest;
   evidence?: InstallTransactionEvidence;
-  /** For library installs: results for each artifact */
+  /**
+   * For library installs: per-artifact result (backward-compat shape).
+   *
+   * Retained as `InstallResult[]` so existing callers/tests keep reading
+   * `.success` / `.name` / `.version`. The authoritative per-artifact STATE
+   * (skipped / success / failed / rolled_back) lives in `journal` (arc#227).
+   */
   artifacts?: InstallResult[];
+  /**
+   * For library installs (arc#227 / F-6c): the full transactional journal —
+   * ordered per-artifact state, errors, and landed-artifact evidence. Present
+   * on every library install (success or failure), absent for standalone.
+   */
+  journal?: LibraryInstallJournal;
 }
 
 /**
@@ -535,7 +562,10 @@ async function installLibrary(
     return { success: false, error: errorMessage(err) };
   }
 
-  // Filter to specific artifact if requested
+  // Filter to specific artifact if requested. A single-artifact install keeps
+  // the original semantics — no ordering / atomic-rollback applies to a set of
+  // one — but still flows through the ordered path below (toposort of one
+  // element is itself).
   if (opts.artifactName) {
     const match = artifactEntries.find(
       (a) => a.manifest.name === opts.artifactName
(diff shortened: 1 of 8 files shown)


src/lib/install-transaction.ts
```diff
@@ -38,6 +38,135 @@ export interface InstallTransaction {
...
+    async rollback() {
+      // Reverse order: last landed artifact unwinds first.
+      for (let i = landed.length - 1; i >= 0; i--) {
+        const { name, tx } = landed[i];
+        await tx.rollback();
+        // The sub-transaction's own rollback does NOT remove the committed DB
+        // row (the DB commit is the last step of a single-artifact install, so
+        // it never had to). The library level removes it here.
+        if (opts.removeDbRow) {
+          opts.removeDbRow(name);
```
mellanon: **[warning/quality]** `removeDbRow` is called bare inside the rollback loop with no error boundary. If `removeSkill(db, name)` throws (e.g. SQLite `SQLITE_BUSY`, locked DB, schema error), the `for` loop throws, the `detail.state` for the remaining artifacts is never updated to `ROLLED_BACK`, and all earlier-in-sequence artifacts are left with un-removed filesystem symlinks/hooks/launchd entries — the exact state the rollback was designed to prevent.

Suggested fix: wrap the `removeDbRow` call in a try/catch that warns into the journal (matching the pattern `beginInstallTransaction.rollback()` uses for every other step):
```typescript
try {
  if (opts.removeDbRow) opts.removeDbRow(name);
} catch (err) {
  // DB removal warning — filesystem was already unwound; log and continue
  const detail = find(name);
  if (detail) detail.error = `DB row removal failed: ${errorMessage(err)}`;
}
```
The sub-transaction `rollback()` already tolerates errors via `warn()` (see `beginInstallTransaction`); `removeDbRow` should get the same treatment so one DB hiccup doesn't strand the unwind loop.

src/commands/install.ts
```diff
@@ -575,44 +667,124 @@ async function installLibrary(
...
+      tx.recordArtifactSuccess(
+        artifactManifest.name,
+        artifactTx,
+        artifactManifest.version,
+        artifactManifest.type,
+      );
+    } else {
+      // Defensive: a success without a captured transaction means nothing for
+      // us to unwind — record state but with no rollback handle.
+      tx.recordArtifactSkipped(
```
mellanon: **[nit/quality]** The `else` branch here (lines 717–723) is unreachable dead code. `onTransaction` fires at line 850 of `installSingleArtifact`, which is *before* any code path that returns `success: true` (all early returns prior to line 850 return `success: false`). So if `artifactResult.success === true`, `artifactTx` is always set — the `else` arm that calls `recordArtifactSkipped` on a successful install can never execute.

The comment "a success without a captured transaction means nothing for us to unwind" accurately describes a theoretical case but it cannot happen with the current `installSingleArtifact` implementation. Options:
- Remove the `else` branch and the `let artifactTx: InstallTransaction | undefined` + replace with `let artifactTx!: InstallTransaction` (asserted non-null after `artifactResult.success`) to make the invariant explicit.
- Or keep as a defensive guard but add a comment clarifying it is currently unreachable (`// unreachable: onTransaction always fires before any success path`).

src/commands/install.ts
```diff
@@ -550,23 +580,85 @@ async function installLibrary(
...
+    return {
+      success: false,
+      name: libraryName,
+      version: libraryManifest.version,
+      error: `Cannot order artifacts of library '${libraryName}': ${errorMessage(err)}`,
+    };
+  }
+
+  // arc#227 / F-6c: resume a failed install from a named artifact. Everything
+  // ordered before it is assumed already installed (or deliberately skipped).
```
mellanon: **[nit/quality]** `resumeFromArtifact` skips everything before `startIndex` with no verification that those predecessors are actually installed. If the user runs `arc install dev-loop --resume-from=pilot` before `agent-state` has landed, `pilot` installs against an unresolved dependency and a subsequent failure rolls back only what landed in THIS run — `agent-state`'s absence is never detected.

This is consistent with the PR description ("assumed already installed or deliberately skipped") and the test at `library-install-ordering.test.ts:131` deliberately pre-installs predecessors before calling resume. The current behavior is the right call for a first-pass implementation.

**Suggested addition**: a follow-up issue documenting this as a known gap ("resume --from=X does not verify X's intra-library predecessors are installed") and noting that a future `--verify-deps` flag could add the DB check. That way the limitation is visible to operators who hand-edit the resume point.

mellanon (commented):
Lenses applied: CodeQuality, Architecture. 2 findings (1 warning, 1 nit) + 1 advisory nit.

**Rollback completeness assessment (adversarial focus):**

The rollback model is correct for the core case: landed artifact sub-transactions are held in `landed[]` in insertion order; rollback iterates in reverse; `removeDbRow` fires after each `tx.rollback()` (DB was the last thing committed so it's the first thing un-committed on reversal). Skipped artifacts are never in `landed[]` so they are provably never touched by rollback. The FAILED artifact is in `details[]` but never in `landed[]`, so it is not double-rolled-back. Single-artifact self-rollback (hooks gate, postinstall failure) happens inside `installSingleArtifact` before success returns; `onTransaction` fires before that and the sub-tx is only added to `landed[]` on the SUCCESS recording path — no double-rollback possible.

One gap: `removeDbRow` (line 153 of install-transaction.ts) has no try/catch. The sub-transaction `rollback()` wraps all other teardown in try/catch + warns. `removeDbRow` does not. A SQLite error mid-loop aborts the `for` loop and leaves earlier-sequence artifacts with their filesystem state unreversed. This is the warning-level finding.

**Toposort:** Kahn's + cycle detection is correct. Independent peer ordering is deterministic (declaration order via stable queue seeding). Self-cycles and direct cycles throw, transitive cycles are caught by the "emitted set vs input length" check. External deps are correctly ignored as ordering edges. ✅

**Resume:** Correctly re-enters at `startIndex` in the sorted order, so it respects the dependency graph. The "already installed" existing-check inside the loop handles idempotency when the named artifact itself is already installed. Does not verify predecessors are actually installed — documented assumption, correct for first-pass.

**Divergences from spec:** All 4 are defensible. Divergence 3 (live-handle vs evidence-replay) is the right call — it avoids re-deriving state and reuses the proven `beginInstallTransaction.rollback()` path. Process-crash-during-rollback is an acknowledged documentation-only gap (crash = no unwind; journal is in-memory).

**arc#230 types.ts conflict:** Non-conflicting. arc#230 edits ArtifactType line 6 + adds process types at line 265; arc#231 adds re-exports at line 402. Different regions, clean 3-way merge regardless of merge order.

verdict: blockers=0 majors=1 nits=2 — recommend: request-changes. Fix the `removeDbRow` error boundary (warning), then this is mergeable. (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #233, opened 2026-06-10
feat(install): F-6b — agent identity provisioning at install

mellanon:
Closes #228
Refs the-metafactory/cortex#835 (dev-loop epic, F-6b slice; design §6.2/§6.4)

## What

Provision an agent instance's **NKey seed + DID + state scaffold** automatically when `arc install` lands a `type: agent` package — no manual post-install steps. Closes the F-6b gap in the dev-loop blueprint.

## Provisioning flow (`src/lib/identity-provision.ts`)

1. **Resolve agent id** — `MF_AGENT_ID` env → `manifest.identity.id` → slug of `manifest.name`. Grammar-guarded (`^[a-z](?:[a-z0-9]|-(?=[a-z0-9]))*$`).
2. **NKey seed** at `~/.config/nats/<agent-id>.nk`, **chmod 600**. Generator chain: `nsc generate nkey -u` → `nkeys.js` → fail-closed guidance. Mirrors cortex `scripts/lib/stack-identity-provision.sh`.
3. **DID** = `did:mf:<agent-id>` (no principal segment; mechanical). Recorded in instance metadata.
4. **Best-effort pubkey** derivation (empty acceptable — cortex logs at boot).
5. **Instance-state scaffold** at `~/.config/cortex/agents/<agent-id>/` — `state.sqlite`, `dashboard.md`, `CLAUDE.md`, `context/{repos,channels}.md`, `retros/`. Mirrors agent-state `ScaffoldFolders` (no hard dep on that repo).
6. **Metadata** (`provisioned=1`, `provisioned_at`, `did`, `nkey_seed_path`, `nkey_pub`) UPSERTed into `state.sqlite`.

Wired into `install.ts` as a **single, clearly-commented hook** (`maybeProvisionAgentIdentity`) at the identity step of both the standalone and library-artifact install paths.

## Storage conventions

| Artifact | Path | Mode |
|---|---|---|
| NKey seed | `~/.config/nats/<agent-id>.nk` | 600 |
| Instance state | `~/.config/cortex/agents/<agent-id>/` | dir |

Env contract: `MF_AGENT_ID` / `MF_INSTANCE_DIR` / `MF_NATS_DIR` / `MF_PRINCIPAL` (the last is correlation-only, never identity).

## Fail-closed rules (cortex#563 precedent)

- **Rule 1 — no identity without a state skeleton.** If the instance dir can't be created (e.g. EACCES), WARN + skip identity wiring; **no orphan seed**. Install still succeeds; agent boots unidentified.
- **Rule 2 — idempotent.** Existing seed reused (never regenerated); operator-edited files never overwritten; partial state scaffolds only the missing pieces. Survives `arc upgrade`.
- **Never throws** — every failure path returns `{ provisioned: false, warning }`.
- **Invalid agent id** is refused rather than written to a bad path.

## Tests

**15 new tests, all green** (12 unit in `test/unit/identity-provision.test.ts` + 3 install e2e in `test/commands/install-identity-f6b.test.ts`): DID derivation, path resolution, happy-path NKey/DID/scaffold/metadata, idempotency (seed reuse + operator-file preservation), fail-closed (bad id, uncreatable dir → no orphan), install-hook gating (non-agent no-op), env overrides (`MF_AGENT_ID`/`MF_INSTANCE_DIR`), name-slug fallback, full `install()` flow + non-agent negative.

## Gates

- `bunx tsc --noEmit` ✅
- `bun run lint:errors` ✅ (eslint --quiet, clean)
- `bun test` ✅ — full suite 1044 pass / 0 fail / 3 skip (the task-noted e2e git-commit failures did not occur in this environment; baseline was fully green)
- CI is the source of truth for green per the arc REVIEW_REQUIRED gate.

## Merge-coordination note

This lane keeps all logic in a **new dedicated module** (`src/lib/identity-provision.ts`) and a **single hook call** at the identity step — deliberately **non-adjacent** to the F-6c (library ordering, `install-transaction.ts`) and F-6e (secret provisioning) insertion points, to minimize batch-merge conflict across the concurrent arc install lanes (all off origin/main, behind the REVIEW_REQUIRED gate). The install orchestration is **not** broadly refactored.

### DIFF
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -44,6 +44,11 @@ import {
   beginInstallTransaction,
   type InstallTransactionEvidence,
 } from "../lib/install-transaction.js";
+// F-6b (arc#228): agent identity provisioning. Lives in its own module; wired
+// in below as a SINGLE hook call at the identity step (merge-coordination with
+// the F-6c / F-6e install lanes — keep this concern isolated and its insertion
+// point non-adjacent to theirs).
+import { maybeProvisionAgentIdentity } from "../lib/identity-provision.js";
 
 export interface InstallOptions {
   /** arc's own state paths (configRoot, dbPath, reposDir, …). Host-independent. */
@@ -497,6 +502,14 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
   );
   tx.recordDbCommit(manifest.name);
 
+  // F-6b (arc#228) — IDENTITY STEP. For type:agent packages, provision the
+  // agent's NKey seed + DID and scaffold its instance state. Best-effort and
+  // fail-closed (cortex#563): on any guard trip this WARNs and returns without
+  // throwing, so the install still succeeds and the agent boots unidentified
+  // until the operator closes the gap. The SECRETS step (F-6e) owns a separate,
+  // non-adjacent hook; LIBRARY ORDERING (F-6c) lives in install-transaction.ts.
+  await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
+
   return {
     success: true,
     name: manifest.name,
@@ -764,6 +777,12 @@ export async function installSingleArtifact(
     manifest,
   );
   tx.recordDbCommit(manifest.name);
+
+  // F-6b (arc#228) — IDENTITY STEP (library-artifact path). dev-loop ships its
+  // agents as library artifacts (design §6.1), so the per-artifact install must
+  // also provision identity. Same fail-closed, best-effort hook as the
+  // standalone path above.
+  await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
 
   return {
     success: true,
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -48,7 +48,10 @@ import {
 // in below as a SINGLE hook call at the identity step (merge-coordination with
 // the F-6c / F-6e install lanes — keep this concern isolated and its insertion
 // point non-adjacent to theirs).
-import { maybeProvisionAgentIdentity } from "../lib/identity-provision.js";
+import {
+  maybeProvisionAgentIdentity,
+  reportProvisioningResult,
+} from "../lib/identity-provision.js";
 
 export interface InstallOptions {
   /** arc's own state paths (configRoot, dbPath, reposDir, …). Host-independent. */
@@ -508,7 +511,10 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
   // throwing, so the install still succeeds and the agent boots unidentified
   // until the operator closes the gap. The SECRETS step (F-6e) owns a separate,
   // non-adjacent hook; LIBRARY ORDERING (F-6c) lives in install-transaction.ts.
-  await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
+  // A fail-closed/skip outcome is surfaced UNCONDITIONALLY (even under --yes) so
+  // a non-interactive install never hides an unidentified-agent gap.
+  const identityResult = await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
+  reportProvisioningResult(identityResult);
 
   return {
     success: true,
@@ -781,8 +787,9 @@ export async function installSingleArtifact(
   // F-6b (arc#228) — IDENTITY STEP (library-artifact path). dev-loop ships its
   // agents as library artifacts (design §6.1), so the per-artifact install must
   // also provision identity. Same fail-closed, best-effort hook as the
-  // standalone path above.
-  await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
+  // standalone path above — and the same unconditional failure-visibility rule.
+  const artifactIdentityResult = await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
+  reportProvisioningResult(artifactIdentityResult);
 
   return {
     success: true,
--- a/src/lib/identity-provision.ts
+++ b/src/lib/identity-provision.ts
@@ -1,0 +1,628 @@
+/**
+ * identity-provision.ts — agent identity provisioning at install (arc#228, F-6b).
+ *
+ * When an `arc install` lands a `type: agent` package, the agent instance needs
+ * three things provisioned without manual post-install steps:
+ *
+ *   1. an NKey seed at the canonical NATS path (`~/.config/nats/<agent-id>.nk`,
+ *      chmod 600) — the signing identity the agent's daemon binds to;
+ *   2. a DID (`did:mf:<agent-id>`) — the agent's stable wire address; and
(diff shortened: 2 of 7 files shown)


src/commands/install.ts
```diff
@@ -497,6 +502,14 @@ export async function install(opts: InstallOptions): Promise<InstallResult> {
   );
   tx.recordDbCommit(manifest.name);
 
+  // F-6b (arc#228) — IDENTITY STEP. For type:agent packages, provision the
+  // agent's NKey seed + DID and scaffold its instance state. Best-effort and
+  // fail-closed (cortex#563): on any guard trip this WARNs and returns without
+  // throwing, so the install still succeeds and the agent boots unidentified
+  // until the operator closes the gap. The SECRETS step (F-6e) owns a separate,
+  // non-adjacent hook; LIBRARY ORDERING (F-6c) lives in install-transaction.ts.
+  await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
```
mellanon: **[major/security]** Provisioning warnings are silently dropped when `opts.yes=true` (non-interactive/CI installs).

`maybeProvisionAgentIdentity` returns a `ProvisionIdentityResult | null` that carries a `.warning` field when a fail-closed guard fires. Here the return value is discarded entirely — no assignment, no check. Combined with `{ quiet: opts.yes }`, the `record("warn", ...)` calls inside the hook write to nothing when `opts.yes=true`.

**Impact:** A failed provisioning (e.g. `EACCES` on the instance dir, or a malformed agent-id from `MF_AGENT_ID`) produces zero user-visible output in automated installs. The agent silently boots unidentified with no trace in the install log.

**Fix:**
```ts
const provisionResult = await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
if (provisionResult?.warning) {
  process.stderr.write(`[arc install] agent identity provisioning: ${provisionResult.warning}\n`);
}
```
Warnings should always surface to stderr regardless of `quiet` — operators need to know when the agent boots unidentified. Alternatively, add `provisioned`, `provisioningWarning` fields to `InstallResult` so callers can gate on them.

src/commands/install.ts
```diff
@@ -765,6 +778,12 @@ export async function installSingleArtifact(
   );
   tx.recordDbCommit(manifest.name);
 
+  // F-6b (arc#228) — IDENTITY STEP (library-artifact path). dev-loop ships its
+  // agents as library artifacts (design §6.1), so the per-artifact install must
+  // also provision identity. Same fail-closed, best-effort hook as the
+  // standalone path above.
+  await maybeProvisionAgentIdentity(manifest, { quiet: opts.yes });
```
mellanon: **[major/security]** Same warning-suppression issue as line 511 (standalone path): return value discarded, `opts.yes=true` silences all provisioning failure output in the library-artifact install path.

Same fix applies: capture the result and emit `.warning` to stderr unconditionally.

src/lib/identity-provision.ts
```diff
@@ -0,0 +1,628 @@
...
+ * Returns null only when the seed file is missing or malformed (the caller treats
+ * an empty pubkey as acceptable).
+ */
+export async function derivePubkeyFromSeed(seedPath: string): Promise<string | null> {
+  if (!existsSync(seedPath)) return null;
+  try {
+    const { rawSeed } = decodeSeed(readFileSync(seedPath, "utf-8").trim());
+    const pub = await getPublicKeyAsync(rawSeed);
+    return encodePublic(pub, PREFIX_BYTE_USER);
+  } catch {
```
mellanon: **[nit/quality]** Empty `catch` block without `_err` variable naming.

CLAUDE.md rule: *"name the variable with `_err` and add a comment explaining why it's safe to ignore."*

Current:
```ts
} catch {
  // Malformed seed — empty pubkey is acceptable; cortex logs it at boot.
  return null;
}
```

Fix:
```ts
} catch (_err) {
  // Malformed seed — empty pubkey is acceptable; cortex logs it at boot.
  return null;
}
```

The comment explains the reasoning well. Just needs the `_err` naming to satisfy the convention.

src/lib/identity-provision.ts
```diff
@@ -0,0 +1,628 @@
...
+
+/**
+ * Create state.sqlite with a `metadata` key/value table if it doesn't yet hold
+ * one. We deliberately keep this minimal and additive: AgentState's own
+ * migration 0001 (work_items/events) is applied later by `agent-state scaffold`
+ * or by the agent's first run via openState(); creating an empty DB here is
+ * forward-compatible (schema_migrations bookkeeping makes that idempotent).
+ */
+function ensureStateDb(statePath: string): void {
+  const db = new Database(statePath, { create: true });
```
mellanon, line 535: **[nit/security]** `state.sqlite` is created without an explicit file mode — Bun's `new Database(path, { create: true })` uses the process umask default (typically 0o644, world-readable).

The sqlite file contains `nkey_seed_path` (the filesystem path to the 0o600 seed). Path disclosure is low-impact (the seed file itself is protected), but on shared systems it reveals where an agent's signing key lives.

**Fix:** Set restrictive permissions immediately after creation:
```ts
function ensureStateDb(statePath: string): void {
  const db = new Database(statePath, { create: true });
  try {
    db.run("CREATE TABLE IF NOT EXISTS ...");
  } finally {
    db.close();
  }
  chmodSync(statePath, 0o600); // restrict after creation
}
```
Or set the instance dir itself to 0o700 (`mkdirSync(instanceDir, { recursive: true, mode: 0o700 })`) so all files within are protected at the directory level — matching `identity.ts`'s `ensureKeysDir()` pattern.

src/lib/identity-provision.ts
```diff
@@ -0,0 +1,628 @@
...
+import { join } from "node:path";
+import { homedir } from "node:os";
+import { randomBytes } from "node:crypto";
+import { Database } from "bun:sqlite";
+import { getPublicKeyAsync } from "@noble/ed25519";
+
+/** Canonical agent-id grammar: lowercase alnum + non-leading/non-trailing single hyphens. */
+const AGENT_ID_RE = /^[a-z](?:[a-z0-9]|-(?=[a-z0-9]))*$/;
+
+/**
```
mellanon, line 56: **[nit/duplication]** `formatDisplayName()` is a byte-for-byte duplicate of the same function in `src/commands/identity.ts`. The `AGENT_ID_RE` regex is also identical to `NAMING_RE` in `identity.ts`.

Both files will need to be kept in sync if the naming grammar ever changes. Consider extracting to a shared `src/lib/agent-id.ts` (or similar). Since this is in the F-6b merge-coordination lane, a follow-up issue is fine — just noting the drift.

mellanon (commented):
Security review (OWASP full sweep + adversarial key-material focus) — 2 majors, 4 nits.

verdict: blockers=0 majors=2 nits=4 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #234, opened 2026-06-10
feat(install): F-6e — secret provisioning at install

mellanon:
Closes #229
Refs cortex#835 (dev-loop epic) → phase cortex#869; design §6.4 row F-6e.

## What

Provision a `type: agent` package's declared `capabilities.secrets` at install
time and inject them into the **per-agent environment** — never a brief or a log
(cortex `docs/design-agentic-dev-pipeline.md` §6.2 step 5, §3.5b authority
model). The dev-loop's own `APPROVER_GH_TOKEN` / `CORTEX_DEV_GH_TOKEN` are the
motivating examples.

## Secret flow

`arc install <pkg>` detects `capabilities.secrets` and, per declared name:

- **interactive (default):** secure no-echo prompt; Return skips a secret.
- **`--from-env`:** read each declared secret from the current environment.
- **`--skip-secrets`:** store nothing; the daemon starts and fails at first use
  with a clear message.

A storage failure aborts the install cleanly (no symlinks placed yet); a skip
only WARNs, listing the missing NAMES + the `arc secrets set …` command.

## Storage

- **macOS:** Keychain via the `security` CLI. Service key
  `ai.meta-factory.cortex.<agent>.<NAME>`, account-scoped to the principal's
  username.
- **Linux:** systemd `LoadCredential` resolves the file at daemon-start.
- **Universal fallback:** `~/.config/metafactory/secrets/<agent>/<NAME>`, one
  secret per file, **chmod 600 enforced on every write AND read** (cortex#87
  `enforceChmod600` pattern).

`resolveSecretBackend()` selects native-when-available, else the chmod-600 file
backend.

## Injection

`arc install` injects the agent's stored secrets into the **postinstall child's
env only** — a fresh object scoped to that invocation; arc's own process env is
never mutated, so the secrets are gone when postinstall exits (issue §E "unset
after postinstall"). The plist/systemd render resolves a secret by name at
render / daemon-start time.

## Lifecycle verbs

`arc secrets {list,check,set,rotate,remove} <agent> [<secret>]` — `rotate` is
delete-then-add (no in-place overwrite); `set`/`rotate` support `--from-env`.

## Never-log handling (issue §E)

A secret **value** never reaches stdout / stderr / an argv arc logs / an audit
line. Every verb, warning, and error prints NAMES only; the only ingress is the
prompt / env read, the only egress the storage backend and the per-child env.
`redactSecret()` is the single string a diagnostic may print in a value's place
(`(secret redacted)`). Error rethrows are name-scoped and chain `{ cause }`
(whose message is also name-scoped). `rotate` never overwrites in place;
chmod 600 is re-enforced on read.

## Merge coordination

All logic lives in NEW dedicated modules (`src/lib/secrets.ts`,
`src/lib/secret-provision.ts`, `src/lib/secret-provision-install.ts`,
`src/commands/secrets.ts`). The install.ts wiring is a SINGLE clearly-commented
SECRETS hook in `install()` and `installSingleArtifact()`, placed **non-adjacent**
to the F-6b identity hook (near the return, #233) and the F-6c library-ordering
work (`install-transaction.ts`, #231). Concern: SECRETS only.

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint:errors` — clean
- `bun test` — **1081 pass / 0 fail / 3 skip** (52 new tests across 4 files:
  storage backends, flow, install bridge, CLI verbs)

Note: I did NOT see the predicted ~139 e2e git-commit failures in this sandbox —
the full suite ran green. The one e2e failure I hit during development
(`_JIRA` agent-name validation) was mine and is fixed; CI remains the source of
truth.

### DIFF
--- a/src/lib/secrets.ts
+++ b/src/lib/secrets.ts
@@ -1,0 +1,352 @@
+/**
+ * F-6e (arc#229) — secret storage backends.
+ *
+ * Manifests declare `capabilities.secrets: [GITHUB_TOKEN, APPROVER_GH_TOKEN, …]`.
+ * This module owns the *storage* half: where a provisioned secret lives, how it
+ * is retrieved at daemon-start / postinstall time, and how it is rotated and
+ * removed. The install-time *flow* (prompt / --from-env / --skip-secrets) lives
+ * in `secret-provision.ts`; the CLI verbs in `commands/secrets.ts`.
+ *
+ * Backends (issue §B):
+ *   - KeychainBackend — macOS, wraps the `security` CLI. Service-scoped to the
+ *     agent + secret name, account-scoped to the principal's username.
+ *   - FileBackend — universal fallback. One secret per file under
+ *     `<secretsRoot>/<agent>/<NAME>`, chmod 600 enforced on every write AND read
+ *     (cortex#87 `enforceChmod600` pattern).
+ *   - SystemdCredentialsBackend (Linux ≥256) is resolved at *daemon-start* by the
+ *     unit's `LoadCredential` directive, not by arc at install time — arc stores
+ *     to the FileBackend on Linux and the systemd unit's resolver reads it. A
+ *     dedicated read-only backend is deferred (issue "Future considerations").
+ *
+ * NEVER-LOG invariant (issue §E): a secret VALUE must never be written to
+ * stdout / stderr / argv-we-log / an audit line. Backends accept and return
+ * values but never log them. `redactSecret()` is the only string a diagnostic
+ * may print in a value's place. Service keys and file paths are built from the
+ * agent + secret NAME only — never the value.
+ */
+
+import { mkdir, readFile, writeFile, chmod, stat, unlink, readdir } from "fs/promises";
+import { existsSync } from "fs";
+import { join } from "path";
+import { errorMessage, isErrno } from "./errors.js";
+
+/** Fixed sentinel printed in place of a secret value in any diagnostic. */
+export const SECRET_REDACTION = "(secret redacted)";
+
+/**
+ * Return the redaction sentinel. Exists as a named function (rather than the
+ * bare constant) so call sites read as an explicit, greppable redaction — and
+ * so a value accidentally passed in is swallowed, never echoed.
+ */
+export function redactSecret(_value: string): string {
+  return SECRET_REDACTION;
+}
+
+/**
+ * Derive the platform-stable service key for a secret:
+ *   `ai.meta-factory.cortex.<agent>.<SECRET_NAME>`
+ *
+ * Built from NAMES only (issue §B) — the value is never part of the key, so
+ * the key is safe to log.
+ */
+export function secretServiceKey(agent: string, name: string): string {
+  return `ai.meta-factory.cortex.${agent}.${name}`;
+}
+
+// A secret/agent name is an env-var-shaped identifier; reject anything that
+// could escape the per-agent directory or smuggle a separator into a service
+// key. This is the storage-layer guard; the CLI/provision layer validates the
+// manifest declaration up front too.
+const NAME_RE = /^[A-Za-z_][A-Za-z0-9_]*$/;
+
+function assertSecretName(name: string): void {
+  if (!NAME_RE.test(name)) {
+    throw new Error(
+      `invalid secret name "${name}": expected an env-var-shaped identifier ([A-Za-z_][A-Za-z0-9_]*)`,
+    );
+  }
+}
+
+function assertAgentName(agent: string): void {
+  // The agent scope is the manifest package name. Ecosystem packages may lead
+  // with an underscore (e.g. `_JIRA`, `_CONFLUENCE`) and use hyphens
+  // (`dev-loop`). Allow `[A-Za-z0-9_]` to start, then `[A-Za-z0-9_-]` — but
+  // never a path separator or `..` (those would escape `<root>/<agent>/`).
+  if (!/^[A-Za-z0-9_][A-Za-z0-9_-]*$/.test(agent)) {
+    throw new Error(
+      `invalid agent name "${agent}": expected a package-name slug ([A-Za-z0-9_][A-Za-z0-9_-]*)`,
+    );
+  }
+}
+
+/**
+ * A single secret-storage backend. Names in / values in-and-out; the backend
+ * never logs the value.
+ */
+export interface SecretBackend {
+  /** Persist `value` under `name`. Overwrites unless the backend forbids it. */
+  store(name: string, value: string): Promise<void>;
+  /** Return the stored value, or `null` if not present. */
+  retrieve(name: string): Promise<string | null>;
+  /** Delete the secret. Idempotent — no throw if it was already absent. */
+  remove(name: string): Promise<void>;
+  /** Stored secret NAMES for this agent (never values). */
+  list(): Promise<string[]>;
+  /**
+   * Replace the value with no in-place overwrite (issue §E): delete the old
+   * entry, then add the new one. Default impl is remove-then-store; a backend
+   * may override if its native primitive differs.
+   */
+  rotate(name: string, value: string): Promise<void>;
+}
+
+/** Result of a `security` CLI invocation. Mirrors the subset we consume. */
+export interface SecurityResult {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}
+
+/**
+ * Runs the macOS `security` CLI. Injectable so tests never touch the real
+ * login keychain. The default runner spawns `security` synchronously and
+ * captures stdout/stderr (never logged by this module).
+ */
+export type SecurityRunner = (args: string[]) => SecurityResult;
+
+const defaultSecurityRunner: SecurityRunner = (args) => {
+  const result = Bun.spawnSync(["security", ...args], {
+    stdout: "pipe",
+    stderr: "pipe",
+  });
+  return {
+    exitCode: result.exitCode,
+    stdout: result.stdout.toString(),
+    stderr: result.stderr.toString(),
+  };
+};
+
+// `security` returns exit 44 (errSecItemNotFound) when an item is absent.
+const SEC_ITEM_NOT_FOUND = 44;
+
+/**
+ * macOS Keychain backend. Each secret is a generic password keyed by
+ * `secretServiceKey(agent, name)` and scoped to the principal's username.
+ *
+ * The VALUE is passed to `security` via `-w <value>` through the injected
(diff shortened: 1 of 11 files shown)


src/lib/secrets.ts
```diff
@@ -0,0 +1,352 @@
...
+  async store(name: string, value: string): Promise<void> {
+    assertSecretName(name);
+    const service = secretServiceKey(this.agent, name);
+    // -U updates in place if present; for rotate we delete-then-add explicitly.
+    const r = this.runner([
+      "add-generic-password",
+      "-s", service,
+      "-a", this.account,
+      "-U",
+      "-w", value,
```
mellanon, line 176: **[major/security]** `security add-generic-password -w \<value>` places the secret value in the spawned process's `argv[]`, which is world-visible via `ps aux` / `/proc/PID/cmdline` for the duration of the synchronous `spawnSync` call.

**OWASP:** A03 – Sensitive Data Exposure

The comment acknowledges this and draws an analogy to `gh auth login` — but that analogy is misleading: `gh auth login --with-token` reads via **stdin**, not argv. The only option the `security` CLI exposes for the password is `-w` (argv) — no stdin path exists — so this is an inherent limitation of the CLI surface, not a design choice.

**Impact:** On a single-user developer machine the window is extremely brief (`spawnSync` is synchronous; the child process exits immediately). On a shared POSIX box (multi-user CI agent, Docker container with a host PID-namespace mount) any user running `ps aux` in the right millisecond can read the token.

**Fix options (in priority order):**
1. Preferred: replace the `security` CLI with a direct Keychain API call via a small native binding (e.g. `keytar` or a Bun FFI shim to `SecKeychainAddGenericPassword`). Value never enters argv.
2. Acceptable for the declared single-user-macOS scope: add an explicit `@security` warn in the docs noting this is not safe on shared machines, and gate `KeychainBackend` selection on a positive `is single-user machine` check (or just always fall through to `FileBackend` in non-dev contexts).
3. At minimum: update the code comment to say `gh auth login --with-token` uses **stdin**, not argv, so the analogy is imprecise — and explicitly label the argv exposure as the trade-off for avoiding a native binding dep.

src/lib/secrets.ts
```diff
@@ -0,0 +1,352 @@
...
+  private pathFor(name: string): string {
+    assertSecretName(name);
+    return join(this.agentDir, name);
+  }
+
+  async store(name: string, value: string): Promise<void> {
+    const path = this.pathFor(name);
+    await mkdir(this.agentDir, { recursive: true });
+    // Write with mode 600 from the start, then re-chmod in case the file
+    // pre-existed with looser perms (write-with-mode only applies on create).
```
mellanon: **[nit/security]** `FileBackend.store` has a brief TOCTOU window when overwriting an **existing** secret file.

**OWASP:** A03 – Sensitive Data Exposure

`writeFile(path, value, { mode: 0o600 })` — the `mode` option is passed to the underlying `open(2)` call as the creation mode. On POSIX, for an **already-existing** file, `open(O_WRONLY | O_TRUNC)` does NOT apply the mode bits — it opens with the existing file's permissions. So if a secret file somehow drifted to `0o644` before a `store` call, the new value would be written at `0o644` permissions, then `chmod` would tighten it to `0o600`. Between those two syscalls, the file holds the new value at the too-open mode.

**Practical impact:** Low in practice — `rotate` correctly calls `remove` first, so the rotate path never has an existing file. The window only exists on a direct `set` / `provisionSecrets` call on an already-stored secret. On a single-user dev machine this is nearly theoretical.

**Fix:** Write to a temp file in the same directory with mode `0o600`, then atomically rename into place. This eliminates both the TOCTOU and any possibility of partial writes being read:
```ts
const tmp = `${path}.tmp`;
await writeFile(tmp, value, { mode: 0o600 });
await chmod(tmp, 0o600);   // enforce before rename
await rename(tmp, path);   // atomic on same-filesystem
```

src/lib/secrets.ts
```diff
@@ -0,0 +1,352 @@
...
+const SEC_ITEM_NOT_FOUND = 44;
+
+/**
+ * macOS Keychain backend. Each secret is a generic password keyed by
+ * `secretServiceKey(agent, name)` and scoped to the principal's username.
+ *
+ * The VALUE is passed to `security` via `-w <value>` through the injected
+ * runner only; this class never logs argv (issue §E). Note: `security`'s own
+ * `-w` does place the value on the spawned process's argv — that is the OS
+ * keychain tool's contract and is the same path `gh auth login` uses. arc's
```
mellanon: **[nit/quality]** The `gh auth login` analogy in this comment is imprecise.

`gh auth login --with-token` reads the token from **stdin** (`< token.txt` or a pipe), not from a command-line argument. The analogy suggests this is standard CLI practice and implies equivalent exposure; it is not. `gh` specifically avoids argv for this reason.

The honest statement is: the `security` CLI has no stdin path for `-w`; argv is the only option it exposes. That is the actual trade-off — not that argv is a known-safe pattern. Recommend updating to:

> Note: `security`'s own `-w` places the value on the spawned process's argv — this is an inherent limitation of the `security` CLI (no stdin path is available). On a single-user dev machine the window is very brief (`spawnSync` is synchronous). See the `KeychainBackend` architectural note in `docs/secret-provisioning.md` for the native-binding upgrade path.

src/lib/secrets.ts
```diff
@@ -0,0 +1,352 @@
...
+  async rotate(name: string, value: string): Promise<void> {
+    // No in-place overwrite (issue §E): delete the old item, then add the new.
+    await this.remove(name);
+    await this.store(name, value);
+  }
+
+  list(): Promise<string[]> {
+    // `security` has no clean "list items for a service prefix" without
+    // dumping the whole keychain. The authoritative roster of an agent's
+    // secrets is the manifest's `capabilities.secrets`; `list`/`check` resolve
```
mellanon: **[nit/quality]** `KeychainBackend.list()` always returns `[]`, so `arc secrets list <agent>` will silently report "No secrets stored" on macOS even when the agent has Keychain-backed secrets. The code comment explains the rationale (no clean prefix query without a full keychain dump), but the user-visible output is misleading.

**Suggested fix:** When the backend is `KeychainBackend`, `secretsList` should iterate over the manifest's declared names and call `retrieve` to check presence — the same pattern `secretsCheck` uses via `validateSecretPresence`. Or, document the behavior in the CLI help text:

```
arc secrets list <agent>   # on macOS Keychain: lists declared names that are stored
```

The current behaviour violates the principle of least surprise for an operator running `arc secrets list` to verify provisioning state on macOS.

src/lib/secrets.ts
```diff
@@ -0,0 +1,352 @@
...
+    if (r.exitCode !== 0) {
+      // Loud, not silent (issue §E): retrieval failure other than not-found is
+      // an error the operator must see — but NEVER includes the value.
+      throw new Error(
+        `keychain retrieve failed for ${service} (exit ${r.exitCode}): ${r.stderr.trim()}. ` +
+          `Re-run \`arc secrets set <agent> ${name}\` to repair.`,
+      );
+    }
+    // `-w` prints the raw value plus a trailing newline.
+    return Promise.resolve(r.stdout.replace(/\n$/, ""));
```
mellanon: **[nit/quality]** `r.stdout.replace(/\n$/, "")` only strips a single trailing newline. If the stored token itself ends with `\n` (unusual but possible for file-pasted values), the trailing newline is silently stripped on retrieval, changing the value.

This is a very low-probability scenario for typical API tokens, but worth noting. The fix is to strip up to one trailing newline unconditionally — which is what the current regex does — but add a comment clarifying that tokens ending in `\n` will be normalised, so operators who store such values know to expect it.

mellanon (commented):
Security review (full OWASP) — F-6e secret provisioning at install (arc#234).

Architecture: 5 modules, 52 tests, 3 storage backends. Never-log invariant
well-structured; redactSecret sentinel + name-scoped errors + NAMES-only
result structs form a coherent defence.

Point 1 — Never-log: PASS with one major caveat. All stdout/stderr/audit
paths carry names only. Error rethrows are name-scoped; { cause } chains
are name-scoped at the backend layer. One exception: KeychainBackend.store
passes the secret value via `security -w <value>` in argv[] — visible in
`ps aux` for the duration of spawnSync. This is inherent to the `security`
CLI (no stdin path exists) and is disclosed in a comment, but the analogy to
`gh auth login` is wrong (gh uses stdin). On a single-user dev machine the
window is very brief; on a shared machine it is a real exposure.

Point 2 — At-rest: PASS with a nit. Keychain scoping is correct
(per-agent, per-secret, account-scoped). FileBackend: chmod 600 on write
AND read enforced. One TOCTOU window exists when store() overwrites an
EXISTING file (writeFile writes at old perms, chmod follows); rotate() is
safe (remove-then-store). Atomic write (tmp-rename) would eliminate this.

Point 3 — Env injection: PASS. buildSecretEnvForInstall returns a fresh
object; arc's own process.env is never mutated. Declared-but-unstored
secrets are OMITTED (not injected as "undefined"). The child inherits
{...process.env, ...secrets} via runScript; no leakage to sibling steps.

Point 4 — Store-failure handling: PASS. Store failure throws → caught
in installTimeProvisionSecrets → { success: false } → install aborts
cleanly (rm -rf installPath). Skip only warns with names. Fail-closed.

Point 5 — security CLI shell-safety: PASS. All security invocations use
Bun.spawnSync arg arrays (no shell string interpolation). assertSecretName
and assertAgentName guard both the file path join and the Keychain service
key before any values reach those surfaces.

Point 6 — rotate = delete-then-add: PASS. Both backends implement
remove-then-store explicitly. The agent-name regex relaxation (_JIRA:
leading underscore allowed) is safe — . and / remain outside the character
class, so path traversal is blocked. KeychainBackend.store uses -U (update
in place) for regular set; rotate explicitly calls remove first. Correct.

Point 7 — Tests honest: PASS. 52 hermetic tests; injected backends; all
5 CLI verbs tested; assertNoValueLeaked assertions on CLI output; rotation
semantics verified (delete-then-add call order); fail-closed store failure
tested. Minor gaps: no TOCTOU test for existing-file overwrite; no test
for quiet=false skip-warning display; trailing-newline value round-trip.

verdict: blockers=0 majors=1 nits=4 — recommend: request-changes.
The argv exposure on Keychain store is the make-or-break point 1 item;
it needs either a native binding upgrade path filed as a tracked issue,
or an explicit scope restriction in the docs marking KeychainBackend
as unsafe on shared/multi-user machines, plus correction of the
misleading gh-auth-login analogy. (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #237, opened 2026-06-11
feat(arc): F-6a — invoke cortex config merge on install (cortex_config manifest field)

mellanon:
## Arc side of F-6a (cortex#858)

Automated, validated, idempotent merge of a package's declared **cortex** capabilities/policy into a cortex stack's config at install time. The cortex `config merge` verb already exists and is merged (`the-metafactory/cortex` `src/cli/cortex/commands/config-merge.ts`). This PR adds the **arc side**: arc now declares `cortex_config` in the manifest and invokes the verb on install.

> Arc side of cortex#858 (the umbrella issue lives in the cortex repo).

### What shipped

| Area | Change |
|---|---|
| `src/types.ts` | Optional `cortex_config` field on `ArcManifest` — inline (`capabilities`/`policy` only) **or** a `{ path }` pointer to a YAML fragment shipped in the package. |
| `src/lib/manifest.ts` | `validateCortexConfig` structural pre-filter at the manifest edge: capability/policy-only boundary (rejects smuggled `agents`/`principal`/`nats` keys), inline-vs-path mutual exclusivity, relative/no-`..` path guard. Wired into all three `readManifest` return paths (library/process/general). |
| `src/lib/cortex-config-provision.ts` | **New** dedicated module (sibling to `identity-provision.ts` / `secret-provision-install.ts`). Detects host-is-cortex (reuses the existing cortex `HostAdapter` `id` + `detect()`), resolves the stack config dir (`host.paths.root`), marshals the fragment to disk, and invokes `cortex config merge --config <dir> --fragment <file> [--stack <id>]` via a swappable spawn runner (test seam gated on `ARC_TEST_MODE`, mirroring `nats-broker.ts`). cortex resolution: `ARC_CORTEX_BIN`/`MF_CORTEX_BIN` → `cortex` on PATH. |
| `src/commands/install.ts` | Single hook at the cortex-config step ("step 6c") on **both** the standalone and library-artifact paths, **after** the post-landing transaction. Fail-closed. |
| `docs/cortex-config-composition.md` | Integration doc (matches the F-6b/F-6e sibling docs). |

### Fail-closed + rollback

A non-zero exit from `cortex config merge` fails the install. The landed state is unwound: the install transaction rolls back symlinks/hooks/extensions/launchd, and because the DB row is committed as the transaction's **last** step (BEFORE this step runs — every existing rollback path fires before the commit), the cortex-config step removes the row itself. The cortex verb is idempotent and writes a 0o600 backup, so an arc retry after fixing the cause re-runs cleanly.

### Composes with the merged F-6 slices

- **F-6c (#231)** ordered library installs + atomic rollback — `install-transaction.ts`
- **F-6b (#233)** agent identity provisioning — `identity-provision.ts`
- **F-6e (#234)** secret provisioning — `secret-provision-install.ts`

Each slice is a single, clearly-commented, **non-adjacent** hook so the lanes compose without colliding. F-6a captures the live transaction via the same `onTransaction` callback F-6c uses, runs after the identity step, and adds the DB-row removal that the post-commit position requires.

### Tests (40 new, full suite green)

- **Manifest** (`test/unit/manifest-cortex-config.test.ts`): inline accept (caps/policy/both), reject smuggled key, reject empty, reject non-array caps / non-object policy; path accept + absolute/`..`/path+inline rejection; non-object rejection; round-trip through `readManifest`.
- **Provision module** (`test/unit/cortex-config-provision.test.ts`): host detection, cortex-bin resolution, fragment marshalling (inline temp file + path resolution + traversal/missing guards), argv shape, `--stack` presence/absence, fail-closed on non-zero exit, fail-closed on missing CLI, temp-file cleanup.
- **Install integration** (`test/commands/install-cortex-config-f6a.test.ts`): invokes merge when fragment present + host cortex; skips when absent; skips when host non-cortex; fails closed + rolls back the DB row on merge failure; retry succeeds via the idempotent verb.

Gates: `bunx tsc --noEmit` clean · `bun run lint` clean · `bun test` 1205 pass / 0 fail.

Closes arc side of cortex#858.

### DIFF
--- a/docs/cortex-config-composition.md
+++ b/docs/cortex-config-composition.md
@@ -1,0 +1,110 @@
+# Cortex config composition at install (F-6a / cortex#858)
+
+When `arc install` lands a package whose manifest declares a `cortex_config`
+fragment AND the install target host is a cortex stack, arc merges the package's
+declared capabilities/policy into the stack's config automatically — no manual
+editing of `stacks/<id>.yaml`. This closes the F-6a gap in the dev-loop
+blueprint (cortex `docs/design-agentic-dev-pipeline.md` §6.2 install lifecycle
+step 3 / "step 6c", §6.4).
+
+Implementation: [`src/lib/cortex-config-provision.ts`](../src/lib/cortex-config-provision.ts),
+wired into [`src/commands/install.ts`](../src/commands/install.ts) as a single
+hook call (`maybeMergeCortexConfig`) at the cortex-config step of both the
+standalone and library-artifact install paths — AFTER the post-landing
+transaction (so postinstall has run), non-adjacent to the F-6b identity hook and
+the F-6e secrets hook.
+
+## The `cortex_config` manifest field
+
+Optional. The package's CORTEX capability/policy declaration — distinct from
+`capabilities`, which is arc's OWN host-side enforcement (what the package may
+read/write/network/shell locally). Two mutually-exclusive forms:
+
+```yaml
+# Inline form — capabilities and/or policy, NOTHING else.
+cortex_config:
+  capabilities:
+    - id: dev.implement
+      description: "Dev agent implementation"
+      provided_by: [dev-agent]
+  policy:
+    principals:
+      - id: dev-agent
+        role: [develop]
+    roles:
+      - id: develop
+        capabilities: [dev.implement]
+```
+
+```yaml
+# Path-pointer form — a relative path to a YAML fragment shipped in the package.
+cortex_config:
+  path: cortex-config.yaml
+```
+
+A fragment may carry ONLY `capabilities` and/or `policy`. arc rejects any other
+top-level key (`agents`, `principal`, `nats`, …) at manifest-read time so a
+package can't smuggle a transport/identity change in through the merge path —
+that config is the stack's, not the package's. The path form is guarded to a
+relative path with no `..` traversal.
+
+## What runs
+
+arc does NOT reimplement the merge (Anti-Abstraction Gate). It invokes the
+cortex CLI verb, which owns the deep semantics:
+
+```
+cortex config merge --config <stack-config-dir> --fragment <file> [--stack <id>]
+```
+
+| cortex-side behavior | Consequence for arc |
+|---|---|
+| id-keyed deep merge (capabilities/principals/roles), fragment-wins | The package's declarations land in `stacks/<id>.yaml`. |
+| **idempotent** — same fragment twice is a no-op | A retry after a partial failure is safe. |
+| `CortexConfigSchema.parse()` of the composed whole before write | A fragment that would break the stack is rejected; install fails closed. |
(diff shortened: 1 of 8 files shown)


mellanon (commented):
## Self-review (CodeReview skill — 5 lenses, run directly; Agent spawn unavailable in this env)

Reviewed the full diff across CodeQuality, Security, Architecture, EcosystemCompliance, and Performance lenses. No blockers. Findings below.

### 🟢 Verified correct
- **Fail-closed rollback (the flagged concern).** The single-package `InstallTransaction.rollback()` unwinds hooks/extensions/symlinks/launchd but deliberately stops short of the DB commit (its last step — every pre-existing rollback path fires *before* `recordDbCommit`). The F-6a step runs *after* the commit, so it explicitly `removeSkill(db, name)` after `tx.rollback()`. Verified on both paths. No double-removal in the library path: `installSingleArtifact` owns its own full unwind on a failure return, and `installLibrary` only records SUCCESS artifacts as rollback targets — the FAILED artifact is never re-unwound.
- **Capability/policy-only boundary.** Enforced twice: `validateCortexConfig` rejects any non-`capabilities`/`policy` top-level key at manifest-read time, and `marshalFragment` writes *only* `capabilities`/`policy` to the fragment file (so even a hand-constructed fragment can't smuggle transport/identity keys past arc). cortex's `CapabilityMergeFragmentSchema.strict()` is the authoritative third gate.
- **Path-traversal guards.** Double-guarded: manifest validator (relative + no leading `/` + no `..`) and `marshalFragment` (path.relative containment check). Existence checked before invoking the verb.
- **Spawn/test seam.** Mirrors `nats-broker.ts` exactly — `__setRunnerForTests`/`__resetRunnerForTests` gated on `ARC_TEST_MODE`/`NODE_ENV=test`; the unit + integration suites never shell out to a real cortex.
- **Composition with F-6b/F-6c/F-6e.** Single non-adjacent hook; captures the live transaction via the same `onTransaction` callback F-6c uses; runs after the identity step.

### 🔵 Suggestions (non-blocking, addressing inline)
1. **`marshalFragment` empty-fragment path.** If `maybeMergeCortexConfig` is called directly (bypassing the manifest validator) with a fragment that has neither `path` nor `capabilities` nor `policy`, it writes `{}` to the temp file. cortex then rejects it (exit 1 → fail-closed), so it's safe — but a defensive early guard would give a clearer arc-side error. Low value given the install path always goes through the validator; noting for completeness.
2. **`host-not-cortex` is silent under `--yes`.** A package that declares `cortex_config` but is installed onto a non-cortex host skips the merge with a stdout line that's suppressed under `quiet`. This is the correct behavior (the fragment legitimately doesn't apply), but unlike the F-6b identity hook — which surfaces a skip *unconditionally* — this stays quiet. Intentional: a non-applicable fragment is not a gap the operator must close. Documented in the integration doc.

**Verdict:** approve (self-review). Gates green: `tsc` clean, `lint` clean, `bun test` 1205 pass / 0 fail.

PR #238, opened 2026-06-12
feat(install): cortex bot-pack drop — agent.yaml/persona.md install flow (cortex#1021 W-4)

jcfischer:
## What

W-4 of cortex `docs/plan-bot-packs-completion.md` (tracking cortex#1021; design: cortex `docs/design-bot-packs.md` §4, `docs/design-arc-agent-bots.md` §6.2/§8.1).

A `type: agent` pack installed on a **cortex target** whose root carries `agent.yaml` (the **bot-pack shape** — yarrow is the first) now installs as:

| Pack file | Lands at |
|---|---|
| `agent.yaml` | `{configRoot}/agents.d/<id>.yaml` — `<id>` from the fragment's own `id:`, validated as a safe filename stem; a present-but-unsafe id is an install ERROR (with rollback); absent id falls back to the sanitized manifest name |
| `persona.md` | `{configRoot}/personas/<id>.md` (optional) |

**Branch condition is cortex-host AND agent.yaml shape.** Legacy standalone-bot repos (fragment via `provides.files`, sage shape) keep the existing path — pinned by the untouched `install-multitarget-i140` tests. Non-cortex hosts keep the legacy `.md` agent drop.

**Post-install side effects without hardcoding cortex:** the §8.1 ordering (drop fragment → `cortex agents reload` → `cortex creds issue <id>`) rides the pack's `lifecycle.postinstall` scripts, which `install()` already runs after the artifact drop. Test pins reload-before-creds ordering.

**Hardening that fell out of review:** `installPerTarget` now catches a throw from the per-target artifact drop, rolls back already-installed targets, and surfaces a normal install error (previously an uncaught escape with no rollback).

## Tests
`test/commands/install-cortex-botpack.test.ts`: fragment+persona under the fragment id; lifecycle ordering; persona-optional; absent-id fallback; path-traversal id → install error, nothing lands. Full suite: 1212 tests green (one env-contention flake, 6/6 in isolation).

## Follow-up
`arc install yarrow` e2e against the real pack (github.com/the-metafactory/yarrow) once merged.

### DIFF
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -1042,16 +1042,25 @@ async function installPerTarget(opts: {
       };
     }
 
-    // registry hosts (cortex, claude-code) take the existing symlink path
-    const r = await createArtifactSymlinks({
-      type: opts.manifest.type,
-      manifest: opts.manifest,
-      arc: opts.arc,
-      host: targetHost,
-      installDir: opts.installPath,
-      consumerDir: opts.consumerDir,
-      quiet: opts.quiet,
-    });
+    // registry hosts (cortex, claude-code) take the existing symlink path.
+    // A THROW from the artifact drop (e.g. a bot pack refusing an unsafe
+    // fragment id) must roll back the targets already installed and surface
+    // as a normal install error, not an uncaught exception.
+    let r: Awaited<ReturnType<typeof createArtifactSymlinks>>;
+    try {
+      r = await createArtifactSymlinks({
+        type: opts.manifest.type,
+        manifest: opts.manifest,
+        arc: opts.arc,
+        host: targetHost,
+        installDir: opts.installPath,
+        consumerDir: opts.consumerDir,
+        quiet: opts.quiet,
+      });
+    } catch (err) {
+      await rollbackAll();
+      return { error: `[${targetId}] ${errorMessage(err)}` };
+    }
     if (r.filesMissingSource.length) {
       const detail = r.filesMissingSource
         .map((f) => `  - ${f.source} -> ${f.target}`)
--- a/src/lib/artifact-installer.ts
+++ b/src/lib/artifact-installer.ts
@@ -494,11 +494,13 @@ export function toposortArtifacts(artifacts: ArtifactEntry[]): ArtifactEntry[] {
  * not a string, or the YAML is unreadable (cortex's own loader still
  * validates the fragment content after the drop).
  *
- * The §8.1 post-install side effects — `cortex agents reload`, then
- * `cortex creds issue <id>` for daemon brains — are the PACK's
- * `lifecycle.postinstall` scripts, which install() runs after this drop.
- * Drop → reload → creds: the ordering invariant holds without arc
- * hardcoding cortex CLI calls.
+ * Post-install side effects: arc guarantees ONLY that the artifact drop
+ * happens before `lifecycle.postinstall` scripts run, and that those
+ * scripts run in their declared order. The §8.1 sequence (drop fragment →
+ * `cortex agents reload` → `cortex creds issue <id>`) therefore holds IFF
+ * the pack ships those scripts in that order (as yarrow does) — it is a
+ * pack-authored convention arc sequences, not an arc-enforced cortex
+ * side effect. arc itself never invokes cortex CLI calls.
  */
 async function linkCortexBotPackSymlinks(opts: {
   fragmentPath: string;
--- a/src/lib/artifact-installer.ts
+++ b/src/lib/artifact-installer.ts
@@ -217,7 +217,16 @@ export async function createArtifactSymlinks(opts: {
         // drop. Drop → reload → creds: the ordering invariant holds without
         // arc hardcoding cortex CLI calls.
         const cortexPaths = host.paths as CortexHostPaths;
-        const agentId = readAgentFragmentId(botPackFragment) ?? manifest.name.toLowerCase();
+        const agentId =
+          readAgentFragmentId(botPackFragment) ??
+          sanitizeAgentId(manifest.name.toLowerCase());
+        if (agentId === undefined) {
+          throw new Error(
+            `agent pack "${manifest.name}": neither the agent.yaml id nor the ` +
+              `manifest name is a safe filename stem (lowercase alphanumerics ` +
+              `plus . _ -, no path separators) — refusing to drop the fragment`,
+          );
+        }
         await linkTracked(botPackFragment, join(cortexPaths.agentsDir, `${agentId}.yaml`));
         const personaPath = join(installDir, "persona.md");
         if (existsSync(personaPath)) {
@@ -503,7 +512,7 @@ export function readAgentFragmentId(fragmentPath: string): string | undefined {
     const raw = YAML.parse(readFileSync(fragmentPath, "utf-8")) as unknown;
     if (raw === null || typeof raw !== "object" || Array.isArray(raw)) return undefined;
     const id = (raw as Record<string, unknown>).id;
-    return typeof id === "string" && id.trim().length > 0 ? id.trim() : undefined;
+    return typeof id === "string" ? sanitizeAgentId(id.trim()) : undefined;
   } catch (err) {
     // Fall back to the manifest-name stem; the fragment will still fail
     // loudly in cortex's own loader if it is genuinely malformed.
@@ -512,4 +521,20 @@ export function readAgentFragmentId(fragmentPath: string): string | undefined {
     );
     return undefined;
   }
+}
+
+/**
+ * An agent id is used as a FILENAME STEM under agents.d/ and personas/ — it
+ * must not be able to escape those directories (sage arc#238 round 1
+ * blocker: an id containing `/` or `..` would symlink outside agents.d).
+ * Accepts cortex-conventional ids only: lowercase-insensitive alphanumerics
+ * plus `.`, `_`, `-`, starting alphanumeric, and no `..` sequence anywhere.
+ * Returns `undefined` for anything else so the caller can fall back or
+ * refuse loudly.
+ */
+export function sanitizeAgentId(id: string): string | undefined {
+  if (id.length === 0 || id.length > 128) return undefined;
+  if (!/^[a-z0-9][a-z0-9._-]*$/i.test(id)) return undefined;
+  if (id.includes("..")) return undefined;
+  return id;
 }
--- a/src/lib/artifact-installer.ts
+++ b/src/lib/artifact-installer.ts
@@ -197,41 +197,20 @@ export async function createArtifactSymlinks(opts: {
 
     case "agent": {
       const botPackFragment = join(installDir, "agent.yaml");
+      // Bot-pack branch needs BOTH conditions: a cortex host target AND the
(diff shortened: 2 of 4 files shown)


jcfischer (commented):
## Sage code review — changes-requested

1 finding(s): 1 blocker.


### HonestOracle

- **[blocker]** `src/lib/artifact-installer.ts:219` — **Fragment id can escape agentsDir**
  The diff claims `agent.yaml  → {agentsDir}/<id>.yaml`, but the code uses the raw id in `join(cortexPaths.agentsDir, `${agentId}.yaml`)` after accepting any non-empty string id. An `id` containing `/` or `..` can place the symlink outside `agents.d`, so the stated landing contract is not enforced.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
## Sage code review — changes-requested

4 finding(s): 2 important, 1 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/lib/artifact-installer.ts:224` — **Invalid fragment id still gets installed**
  When `readAgentFragmentId(botPackFragment)` returns undefined for an unsafe `id`, the code silently falls back via `sanitizeAgentId(manifest.name.toLowerCase())`, so a fragment containing `id: "../../outside/evil"` is still linked into `agents.d/evil.yaml`. That installs a malformed cortex identity fragment under a different filename, contradicting the stated filename/id authority and likely leaving cortex to reject or mis-handle the pack after arc reports success.
  Fix: Treat an unsafe string `id` as an install error; only fall back when `id` is absent or not a string.

### Maintainability

- **[suggestion]** `src/lib/artifact-installer.ts:199` — **Extract bot-pack symlink branch**
  `if (host.id === "cortex" && existsSync(botPackFragment)) {` now embeds cortex bot-pack id parsing, validation, fragment linking, and persona linking inside the already broad `createArtifactSymlinks` switch case.
  Fix: Extract this branch to `linkCortexBotPackSymlinks(...)` and keep the case as shape detection plus fallback.

- **[nit]** `test/commands/install-cortex-botpack.test.ts:95` — **Deduplicate repo commit setup**
  The fixture repeats `Bun.spawnSync(["git", "add", "."]` and the long `git -c user.name=Test ... commit` block in the helper and mutation tests.
  Fix: Add a small `commitAll(repoDir, message)` test helper.

### HonestOracle

- **[important]** `src/lib/artifact-installer.ts:215` — **Shape-gated claim is false**
  The comment says "Shape-gated on `agent.yaml` presence", and the PR description says "Shape-gated, not host-gated", but the branch condition is `if (host.id === "cortex" && existsSync(botPackFragment))`. That misleads downstream readers about the install contract: the new shape is not sufficient unless the host is cortex.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
## Sage code review — changes-requested

3 finding(s): 1 important, 2 suggestion.


### Maintainability

- **[suggestion]** `src/lib/artifact-installer.ts:198` — **Extract bot-pack symlink branch**
  The branch starting with `if (host.id === "cortex" && existsSync(botPackFragment)) {` mixes shape detection, cortex path casting, id fallback, two symlink destinations, and persona optionality inside `createArtifactSymlinks`. That makes the already broad switch arm harder to scan than an extracted `linkCortexBotPack(...)` helper would be.
  Fix: Move the cortex bot-pack block into a focused helper that returns whether it handled the artifact.

### HonestOracle

- **[important]** `src/lib/artifact-installer.ts:211` — **Shape-gated claim contradicts condition**
  The comment says "Shape-gated on `agent.yaml` presence", but the branch is guarded by `if (host.id === "cortex" && existsSync(botPackFragment))`. That misleads readers into believing shape alone selects this path when host identity is also required.

- **[suggestion]** `test/commands/install-cortex-botpack.test.ts:15` — **Header claims nonexistent local coverage**
  The test header says the legacy shape must "stay on the old path — pinned below", but this file contains no legacy standalone-bot test below it. That points maintainers to coverage that is not present in the added test file.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
## Sage code review — changes-requested

2 finding(s): 2 important.


### HonestOracle

- **[important]** `src/lib/artifact-installer.ts:198` — **Shape-only claim is false**
  The PR claims "Shape-gated, not host-gated", but the branch is explicitly `if (host.id === "cortex" && existsSync(botPackFragment))`. That makes `host.id === "cortex"` load-bearing, so the description misstates the actual gate readers would rely on.

- **[important]** `src/lib/artifact-installer.ts:517` — **Malformed fragment failure is unproven**
  The catch block says "the fragment will still fail loudly in cortex's own loader", but the code only `console.warn`s, returns `undefined`, and lets install fall back to the manifest name. The diff does not prove that later loader failure happens before postinstall side effects, so this is a silent-failure claim sold as enforcement.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
## Sage code review — changes-requested

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/lib/artifact-installer.ts:497` — **Ordering guarantee exceeds evidence**
  The comment claims "Drop → reload → creds: the ordering invariant holds", but this code only drops symlinks; the diff shows fixture lifecycle scripts that write "reload" and "creds", not enforcement that real packs invoke `cortex agents reload` then `cortex creds issue <id>`. That misleads downstream readers into treating pack-authored scripts as an arc-guaranteed cortex side-effect ordering.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
## Sage code review — changes-requested

2 finding(s): 1 important, 1 suggestion.


### HonestOracle

- **[important]** `src/lib/artifact-installer.ts:494` — **Unreadable YAML fallback is overclaimed**
  The comment claims "the YAML is unreadable" may fall back because "cortex's own loader still validates the fragment content after the drop", but this diff only logs a warning and installs under the manifest-derived name. That misleads downstream readers into treating a parse failure as safely deferred validation, while the installer has already created the symlink and can still run postinstall scripts.

- **[suggestion]** `src/lib/hosts/cortex.ts:33` — **Postinstall ordering is not enforced here**
  The comment says "§8.1 post-install side effects (`cortex agents reload`, then `cortex creds issue <id>`) ride the PACK's `lifecycle.postinstall` scripts", but this code does not verify those scripts exist, call those commands, or bind `<id>` to the installed fragment id. It presents pack-provided lifecycle scripts as equivalent to cortex-enforced side effects.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
## Sage code review — changes-requested

2 finding(s): 1 important, 1 suggestion.


### HonestOracle

- **[important]** `src/lib/artifact-installer.ts:508` — **Lifecycle guarantee is overstated**
  The comment claims "Drop → reload → creds: the ordering invariant holds without arc hardcoding cortex CLI calls," but this file only drops symlinks and relies on arbitrary `lifecycle.postinstall` scripts. That misleads readers into believing cortex reload/credential side effects are enforced here when the diff only demonstrates script ordering for a fixture.

- **[suggestion]** `src/lib/artifact-installer.ts:500` — **Unreadable YAML fallback is undocumented policy**
  The comment says "the manifest-name fallback applies only when the id is absent, not a string, or the YAML is unreadable," but the PR description only promises absent-id fallback and unsafe-id rejection. Treating unreadable YAML as a normal fallback is a silent failure mode that is asserted without cited design support.


---
_Posted by Sage on Codex CLI substrate._

jcfischer (commented):
# Review — arc#238 (cortex bot-pack install)

Read the diff against the PR-head tree (d607bfd) and probed `sanitizeAgentId` against a battery of inputs. The shape is solid; comments below are nits, not blockers.

## What I checked
- `sanitizeAgentId` blocks every classic traversal vector I tried — `../foo`, `./foo`, `a/b`, `..foo`, `foo..bar`, leading `.`/`-`, empty, >128 — all return `undefined`. The `.includes("..")` belt-and-suspenders is well placed even though the regex already excludes `/`.
- Throw → rollback wiring in `installPerTarget` correctly catches the new throw site, runs `rollbackAll()` (symlinks AND launchd records), and surfaces a normal install error. The rollback test (`evil.yaml` not created, no `outside/` directory anywhere) pins this end-to-end.
- The bot-pack branch is gated by BOTH `host.id === "cortex"` and `existsSync(agent.yaml)`, so the legacy `.md` path for claude-code hosts and for cortex packs without `agent.yaml` is untouched. The `break` after the new branch correctly falls through to the type-agnostic `provides.files` pass, so a bot pack can still ship aux files.
- The fragment is read from disk after the clone, so even though the symlink later resolves the source dynamically, the id used for the destination filename comes from the verified content at install time. No TOCTOU exposure beyond what arc already accepts for the install dir.
- Post-install ordering claim in the PR description is the right framing — arc only guarantees drop-before-`lifecycle.postinstall` and declared script order; the cortex reload→creds sequence is a pack-authored convention. The cortex.ts header comment matches.

## Findings

### nits

1. **`sanitizeAgentId` accepts trailing `.` / `-` and `.yaml` suffix** (`src/lib/artifact-installer.ts:577`). Probed: `"foo."`, `"foo-"`, and `"yarrow.yaml"` all pass. The last would land as `agents.d/yarrow.yaml.yaml` — cortex's loader presumably catches the stem/id mismatch, but a tighter grammar (`^[a-z0-9][a-z0-9._-]*[a-z0-9]$` plus a `.yaml`/`.yml` suffix reject) would refuse it at the arc boundary and produce a clearer error message. Not security-relevant.

2. **Case preserved from fragment id, lowercased from manifest fallback** (`src/lib/artifact-installer.ts:515` vs `524`). A fragment with `id: Yarrow` lands as `agents.d/Yarrow.yaml`, while a fragment with NO id and manifest name `Yarrow` lands as `agents.d/yarrow.yaml`. Cortex convention is lowercase per the PR description; the two paths producing different cases for "the same" pack is a small surprise. Either lowercase both paths or document that the fragment id is authoritative including case.

3. **`readRawAgentFragmentId` calls `console.warn` unconditionally** (`src/lib/artifact-installer.ts:560`). It has no `quiet` plumbing, so `arc install --yes` (which sets `quiet: true` downstream) still prints the warning on an unreadable fragment. Minor — pass through `quiet` or use the install logger.

4. **`readFileSync` in an async function** (`src/lib/artifact-installer.ts:555`). Works fine for tiny YAML; the rest of this module is async (`fs/promises`). Style consistency only.

5. **Test helper swallows git failures** (`test/commands/install-cortex-botpack.test.ts:46`). `commitAll` ignores `Bun.spawnSync`'s exit code, so a broken setup would proceed with stale state and produce a confusing assertion failure rather than a clear test setup error. Same pattern as existing fixtures, so this is a pre-existing convention call.

## Suggestions (optional, follow-up sized)

- Add a test for a valid uppercase fragment id (`id: Yarrow`) to pin the case-preservation behavior — either as the documented contract or to catch it if someone later normalizes it.
- Consider rejecting `.yaml`/`.yml` suffixes in `sanitizeAgentId` so the destination filename can't end up `<id>.yaml.yaml`.

## Verdict
LGTM as merged. Findings are all nits and can be addressed in a follow-up if desired; the security-critical path (sanitizer + rollback) is correct and well-tested.

